PR #1, opened 2019-09-20
[WIP] create, update and get an incomplete quote

AlexSJ1990:


src/main/kotlin/com/hedvig/underwriter/model/IncompleteQuote.kt
```diff
@@ -0,0 +1,97 @@
...
+@TypeDefs(
+        TypeDef(name = "json", typeClass = JsonStringType::class),
+        TypeDef(name = "jsonb", typeClass = JsonBinaryType::class))
+class IncompleteQuote (
+        @field:Id
+        @field:GeneratedValue(generator = "UUID")
+        @field:GenericGenerator(name = "UUID",strategy = "org.hibernate.id.UUIDGenerator")
+        var id: UUID? = null,
+        var quoteState: QuoteState = QuoteState.INCOMPLETE,
+        var dateStartedRecievingQuoteInfo: Instant,
```
johantjellden: Maybe call this something simpler. `creteationTime` or `createdAt`

src/main/kotlin/com/hedvig/underwriter/model/QuoteInitiatedFrom.kt
```diff
@@ -0,0 +1,7 @@
+package com.hedvig.underwriter.model
+
+enum class QuoteInitiatedFrom {
```
johantjellden: These will probably need to change in the future in order to signal which partner. But lets keep them like this for now.

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/Dtos/QuotePriceResponseDto.kt
```diff
@@ -0,0 +1,6 @@
+package com.hedvig.underwriter.serviceIntegration.productPricing.Dtos
+
+data class QuotePriceResponseDto (
+        val price: Double?
```
AlexSJ1990: I don't want to have a double here, but the pricing model in PP takes a double?

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -68,95 +70,113 @@ class QuoteServiceImpl @Autowired constructor(
...
                     completeQuoteRepository.save(completeQuote)
                     return quotePriceResponseDto
-
-                } catch(exception: Exception) {
-                    throw NullPointerException("Cannot create quote, info missing $exception")
+                } else {
+                    throw NullPointerException("Failed underwriting guideline")
                 }
-            }
+            } catch (exception: Exception) {
```
palmenhq: Can we be a bit more granular in the catch here? Especially considering the above `NullPointerException` will be caught here.

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -68,95 +70,113 @@ class QuoteServiceImpl @Autowired constructor(
...
+                    val quotePriceResponseDto = getQuotePriceDto(completeQuote)!!
+
+                    completeQuote.price = quotePriceResponseDto.price
                     completeQuoteRepository.save(completeQuote)
                     return quotePriceResponseDto
-
-                } catch(exception: Exception) {
-                    throw NullPointerException("Cannot create quote, info missing $exception")
+                } else {
+                    throw NullPointerException("Failed underwriting guideline")
```
palmenhq: Why is this an NPE? Imo we should make some kind of `FailedToUnderwritingGuideline(message: String) : RuntimeException`. Also We could negate the `if(quoteMeetsUWGuidelines)` and throw if its false and then just move on instead of wrapping everything in ifs :)

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -68,95 +70,113 @@ class QuoteServiceImpl @Autowired constructor(
...
                 }
-            }
+            } catch (exception: Exception) {
+            throw NullPointerException("Cannot create quote, info missing $exception")
```
palmenhq: imo we should pass the exception as the cause rather than as a string. That way the stack traces are easier to follow

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -0,0 +1,190 @@
...
+): QuoteService {
+    override fun createIncompleteQuote(incompleteQuoteDto: IncompleteQuoteDto): IncompleteQuoteResponseDto {
+        val incompleteQuote = incompleteQuoteRepository.save(IncompleteQuote.from(incompleteQuoteDto))
+        return IncompleteQuoteResponseDto(incompleteQuote.id!!, incompleteQuote.productType, incompleteQuote.quoteInitiatedFrom)
+    }
+
+    //    TODO: refactor
+    override fun updateIncompleteQuoteData(incompleteQuoteDto: IncompleteQuoteDto, quoteId: UUID) {
+
+        val incompleteQuote = getIncompleteQuote(quoteId)
```
palmenhq: I recently found `org.springframework.beans.BeanUtils`, that guy could probably be helpful when performing this little dance

PR #8, opened 2019-10-10
capture reasons why a quote cannot be completed

AlexSJ1990:


src/main/kotlin/com/hedvig/underwriter/service/QuoteService.kt
```diff
@@ -1,12 +1,12 @@
...
+import arrow.core.Either
 import com.hedvig.underwriter.model.CompleteQuote
 import com.hedvig.underwriter.web.Dtos.*
-import java.time.LocalDate
-import java.time.LocalDateTime
 import java.util.*
+
 interface QuoteService {
-    fun createCompleteQuote(incompleteQuoteId: UUID): CompleteQuoteResponseDto
+    fun createCompleteQuote(incompleteQuoteId: UUID): Either<CompleteQuoteResponseDto, ErrorQuoteResponseDto>
```
johantjellden: The error should be on the left side and the value on the right :)

src/main/kotlin/com/hedvig/underwriter/web/QuoteBuilderController.kt
```diff
@@ -45,7 +46,7 @@ class QuoteBuilderController @Autowired constructor(
     }
 
     @PostMapping("/{id}/completeQuote")
-    fun createCompleteQuote(@Valid @PathVariable id: UUID): ResponseEntity<CompleteQuoteResponseDto> {
+    fun createCompleteQuote(@Valid @PathVariable id: UUID): ResponseEntity<Either<CompleteQuoteResponseDto, ErrorQuoteResponseDto>> {
```
johantjellden: Here as well the left and right should switch places

PR #12, opened 2019-10-15
Major refactoring

johantjellden:


src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -1,98 +1,102 @@
...
                 }
-                else -> completeQuote.startDate = null
+                else -> updatedName.copy(startDate = null)
             }
 
             val memberId = memberService.createMember()
 
-            memberService.updateMemberSsn(memberId!!.toLong(), UpdateSsnRequest(ssn = completeQuote.ssn))
+            if(updatedStartTime.data is PersonPolicyHolder<*>) {
+                memberService.updateMemberSsn(memberId.toLong(), UpdateSsnRequest(ssn = updatedStartTime.data.ssn!!))
```
AlexSJ1990: we can just pass memberId here - have changed within memberService

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -1,98 +1,102 @@
...
+                quote.copy( data = quote.data.updateName(firstName = body.name.firstName, lastName = body.name.lastName))
+            } else { quote }
 
-            when {
+            val updatedStartTime = when {
                 body.startDateWithZone != null -> {
                     val startDateWithZone: DateWithZone = body.startDateWithZone
-                    completeQuote.startDate =
-                            startDateWithZone.date.atStartOfDay().atZone(startDateWithZone.timeZone).toLocalDateTime()
+                    updatedName.copy(startDate =
```
meletios: why startDate is a Datetime? Since we always default to startOfDay is it better to have it as a Date only?

PR #16, opened 2019-10-21
Immutable tables

palmenhq:
📜

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
@@ -1,5 +1,6 @@
 package com.hedvig.underwriter.model
 
+import java.time.Instant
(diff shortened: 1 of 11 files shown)


src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
```diff
@@ -1,111 +1,111 @@
...
     """
     )
-    fun insert(@BindBean quote: DatabaseQuote)
+    @GetGeneratedKeys("id")
+    fun insert(@BindBean quote: DatabaseQuoteRevision, @Bind timestamp: Instant): DatabaseQuoteRevision
+
+    @SqlQuery(
+        """
+            SELECT
+            DISTINCT ON (qr.master_quote_id)
```
johantjellden: Nice :1st_place_medal:

PR #17, opened 2019-10-22
Drop mutable quote tables

palmenhq:
DB lovin ❤️ lets merge this only after we know #16 migrated all data and works like a charm


palmenhq:
@vonElfvin in away i agree, but also we've signed quite a few quotes in dev and i think we should keep them in the underwriter, or?

johantjellden:
Id say that we keep them, lets try to not make DEV more of a mess than it is.

PR #19, opened 2019-10-23
check if member is signed before creating a new member entity

AlexSJ1990:


palmenhq:
Just checking, will this really solve the issue we discussed yesterday/this morning? Considering the member is not added to `signed_member_entity` afai can see the member might onboard through the app after they sign from a partner, hence ending up with duplicate policies no? For that case this wouldn't help (but i might have missed something 😬)

PR #21, opened 2019-10-28
get quote from memberId

AlexSJ1990:


src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
```diff
@@ -79,6 +79,25 @@ interface QuoteDao {
         SELECT * FROM quote_revision_apartment_data WHERE internal_id = :id""")
     fun findApartmentQuoteData(@Bind id: Int): ApartmentData?
 
+    @SqlQuery(
+        """
+            SELECT
+            DISTINCT ON (qr.master_quote_id)
```
meletios: can we have more than one with the same master_quote_id? 🤔

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/dtos/QuoteDto.kt
```diff
@@ -0,0 +1,68 @@
...
+import com.hedvig.underwriter.model.Partner
+import com.hedvig.underwriter.model.ProductType
+import com.hedvig.underwriter.model.QuoteInitiatedFrom
+import com.hedvig.underwriter.model.QuoteState
+import java.math.BigDecimal
+import java.time.Instant
+import java.time.LocalDate
+import java.util.UUID
+
+const val ONE_DAY = 86_400L
```
meletios: Is this in use?

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/dtos/QuoteDto.kt
```diff
@@ -0,0 +1,68 @@
...
+    val currentInsurer: String? = null,
+    val startDate: LocalDate? = null,
+    val validity: Long,
+    val memberId: String? = null,
+    val isComplete: Boolean
+)
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
+@JsonSubTypes(
+    JsonSubTypes.Type(value = ApartmentData::class, name = "home"),
```
meletios:
```suggestion
    JsonSubTypes.Type(value = ApartmentData::class, name = "apartment"),
```

src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
palmenhq: Yes the  revisions have an MtO relationship. ``DISTINCT ON (qr.mqi) ... ORDER BY qr.mqi, qr.id DESC`` selects the most recent one

src/main/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImpl.kt
```diff
@@ -50,6 +51,33 @@ class QuoteRepositoryImpl(private val jdbi: Jdbi) : QuoteRepository {
...
+            state = databaseQuote.state,
+            attributedTo = databaseQuote.attributedTo,
+            currentInsurer = databaseQuote.currentInsurer,
+            startDate = databaseQuote.startDate,
+            price = databaseQuote.price,
+            data = quoteData,
+            memberId = databaseQuote.memberId,
+            initiatedFrom = databaseQuote.initiatedFrom!!,
+            createdAt = databaseQuote.createdAt!!
+        )
```
palmenhq: this looks oddly familiar 🤔 could we maybe extract it to a separate method?

src/main/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImpl.kt
```diff
@@ -50,6 +51,33 @@ class QuoteRepositoryImpl(private val jdbi: Jdbi) : QuoteRepository {
         )
     }
 
+    override fun findByMemberId(memberId: String): Quote? =
```
palmenhq: i think we should make a test of this

src/main/kotlin/com/hedvig/underwriter/web/QuoteController.kt
```diff
@@ -46,6 +47,12 @@ class QuoteController @Autowired constructor(
         return ResponseEntity.ok(optionalQuote)
     }
 
+    @GetMapping("/member/{memberId}")
```
palmenhq: `/members/{memberId}`? :) just to be a bit more "RESTful"

src/main/kotlin/com/hedvig/underwriter/web/QuoteController.kt
```diff
@@ -46,6 +47,12 @@ class QuoteController @Autowired constructor(
...
+    fun getQuoteFromMemberId(@PathVariable memberId: String): ResponseEntity<QuoteDto> {
+        val optionalQuote = quoteService.getQuoteFromMemberId(memberId) ?: return ResponseEntity.notFound().build()
```
palmenhq: technically this is not optional, since we return `notFound` if it's null right?

src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
```diff
@@ -216,6 +217,46 @@ data class Quote(
         return Left(errorStrings)
     }
 
+    fun getQuoteDto(): QuoteDto {
```
palmenhq: could we rename this to `toQuoteDto`? `get` typically means we want to retrieve a property

src/main/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImpl.kt
```diff
@@ -29,10 +30,23 @@ class QuoteRepositoryImpl(private val jdbi: Jdbi) : QuoteRepository {
...
+        val databaseQuote = dao.findByMemberId(memberId) ?: return null
+        return findQuote(databaseQuote, dao)
+    }
+
+    fun findQuote(databaseQuote: DatabaseQuoteRevision, dao: QuoteDao): Quote? {
         val quoteData: QuoteData = when {
             databaseQuote.quoteApartmentDataId != null -> dao.findApartmentQuoteData(databaseQuote.quoteApartmentDataId)
             databaseQuote.quoteHouseDataId != null -> dao.findHouseQuoteData(databaseQuote.quoteHouseDataId)
-            else -> throw IllegalStateException("Quote data must be apartment or house (but was neither) for quote $quoteId}")
+            else -> throw IllegalStateException("Quote data must be apartment or house (but was neither) for quote ${databaseQuote.id}}")
```
palmenhq: is this intensional? Maybe we could log both the quote's id and the revision's id instead?

src/main/kotlin/com/hedvig/underwriter/web/QuoteController.kt
```diff
@@ -46,6 +47,12 @@ class QuoteController @Autowired constructor(
         return ResponseEntity.ok(optionalQuote)
     }
 
+    @GetMapping("/members/{memberId}")
+    fun getQuoteFromMemberId(@PathVariable memberId: String): ResponseEntity<QuoteDto> {
```
palmenhq: sorry for asking so late, but should this really not be a list? What if you have multiple quotes?

src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
```diff
@@ -216,6 +217,46 @@ data class Quote(
         return Left(errorStrings)
     }
 
+    fun toQuoteDto(): QuoteDto {
```
johantjellden: I think that we should move these functions to `QuoteDto` instead. From an architectural standpoint it makes more sens that the `QuoteDto` knows about our `Quote` then that our `Quote` knows about the `QuoteDto` class.
@palmenhq @meletios what do you think?

palmenhq: Sounds reasonable 👍

src/main/kotlin/com/hedvig/underwriter/web/QuoteController.kt
palmenhq: Hm vaild point... Maybe we should filter out the most "suitable" quote, and that should probably be underwriter that does it, right? But i do believe that should be an explicit endpoint so maybe we could rename this one for now?

PR #22, opened 2019-10-30
Modify products + HOUSE 🏠

palmenhq:
dependent on HedvigInsurance/product-pricing#361

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/model/QuoteData.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/QuoteData.kt
@@ -105,7 +105,46 @@ data class HouseData(
 
     override fun updateName(firstName: String, lastName: String) = this.copy(firstName = firstName, lastName = lastName)
 
-    override fun passUwGuidelines() = listOf("You shall not pass!")
+    override fun passUwGuidelines(): List<String> {
+        val errors = mutableListOf<String>()
+
+        if (this.householdSize!! < 1) {
+            errors += "breaches underwriting guideline household size, must be at least 1"
+        }
+        if (this.livingSpace!! < 1) {
+            errors += "breaches underwriting guidline living space, must be at least 1 sqm"
+        }
+
+        if (householdSize!! > 6) {
+            errors += "breaches underwriting guideline household size, must not be more than 6"
+        }
+
+        if (livingSpace!! > 250) {
+            errors += "breaches underwriting guideline living space, must not be more than 250 sqm"
+        }
+
+        if (yearOfConstruction!! < 1925) {
+            errors += "breaches underwriting guideline year of construction, must not be older than 1925"
+        }
+
+        if (numberOfBathrooms!! > 2) {
+            errors += "breaches underwriting guideline number of bathrooms, must not be more than 2"
+        }
+
+        if (extraBuildings!!.filter { building -> building.area > 6 }.size > 4) {
+            errors += "breaches underwriting guideline extra building areas, number of extra buildings with an area over 6 sqm must not be more than 4"
+        }
+
+        if (extraBuildings.any { building -> building.area > 75 }) {
+            errors += "breaches underwriting guideline extra building areas, extra buildings may not be over 75 sqm"
+        }
+
+        if (extraBuildings.any { building -> building.area < 1 }) {
+            errors += "breaches underwriting guideline extra building areas, extra buildings must have an area of at least 1"
+        }
+
+        return errors
+    }
 }
 
 data class ApartmentData(
--- a/src/main/kotlin/com/hedvig/underwriter/model/QuoteData.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/QuoteData.kt
@@ -75,23 +75,23 @@ interface HomeInsurance {
 
 data class HouseData(
     override val id: UUID,
-    override val ssn: String?,
-    override val firstName: String?,
-    override val lastName: String?,
+    override val ssn: String? = null,
+    override val firstName: String? = null,
+    override val lastName: String? = null,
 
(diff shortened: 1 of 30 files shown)


src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -106,11 +113,15 @@ class QuoteServiceImpl(
         return quoteRepository.find(completeQuoteId)
     }
 
-    override fun getQuoteFromMemberId(memberId: String): QuoteDto? {
-        val quote = quoteRepository.findByMemberId(memberId)
-        return QuoteDto.fromQuoteDto(quote!!)
+    override fun getSingleQuoteForMemberId(memberId: String): QuoteDto? {
```
palmenhq: this'll throw an NPE rather than return null if the quote doesn't exist no?

src/main/kotlin/com/hedvig/underwriter/model/QuoteData.kt
```diff
@@ -73,15 +75,24 @@ interface HomeInsurance {
...
+    override val zipCode: String? = null,
+    override val city: String? = null,
+    override var livingSpace: Int? = null,
+    override var householdSize: Int? = null,
+    val ancillaryArea: Int? = null,
+    val yearOfConstruction: Int? = null,
+    val numberOfBathrooms: Int? = null,
+    @Json
+    @get:Json
+    val extraBuildings: List<ExtraBuilding>? = null,
```
johantjellden: I don't think we should persist dto's defined in another package here. We should probably also think about some versioning of them or keep all values nullable in objects that we persist in a json collumn.

src/test/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImplTest.kt
```diff
@@ -192,13 +231,158 @@ class QuoteRepositoryImplTest {
...
+                    ExtraBuilding(
+                        type = "ATTEFALL",
+                        area = 20,
+                        displayName = "Foo",
+                        hasWaterConnected = false
+                    )
+                ),
+                numberOfBathrooms = 2,
+                isSubleted = false
+            ),
```
johantjellden: Next time we touch one of these i propose that we introduce a quote builder.

src/main/kotlin/com/hedvig/underwriter/model/QuoteData.kt
johantjellden: Yeah, i dont know what to think about nullability here. But I think that is it certain that once we introduce a new filed to the objects that are persisted here, we will have ta have a nullable field anyway.

palmenhq: First point, i agree! Will change

Second point i don't agree with. If we want to have loose serialization we should take it as we go imo, introducing nullability because other fields might become nullable in the future is premature imo.

src/test/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImplTest.kt
palmenhq: Haha! Yes

src/main/kotlin/com/hedvig/underwriter/model/QuoteData.kt
palmenhq: regarding the first point, trying to fix it, i don't understand what you mean. `ExtraBuilding` is declared [later in this file[(https://github.com/HedvigInsurance/underwriter/pull/22/files/9135b6a0443b3cb01f6942c2371535b007fba8be#diff-55cb1294d1b4847513723a7ab67f6c7aR211) so i don't understand your point. We're not persisting any dto's as far as i can see.

src/main/kotlin/com/hedvig/underwriter/web/dtos/IncompleteQuoteDto.kt
```diff
@@ -21,7 +28,14 @@ data class IncompleteHouseQuoteDataDto(
...
     val personalNumber: String?,
-    val householdSize: Int?
+    val householdSize: Int?,
+    val ancillaryArea: Int?,
+    val yearOfConstruction: Int?,
+    val numberOfBathrooms: Int?,
+    val extraBuildings: List<ExtraBuildingRequestDto>?,
+    @field:JsonProperty("subleted")
+    val isSubleted: Boolean?,
+    val floor: Int = 0
```
FredrikAreschoug: Is this really needed?

src/main/kotlin/com/hedvig/underwriter/web/dtos/IncompleteQuoteDto.kt
```diff
@@ -21,7 +28,14 @@ data class IncompleteHouseQuoteDataDto(
     val city: String?,
     val livingSpace: Int?,
     val personalNumber: String?,
```
FredrikAreschoug: I know this is not apart of your pr but what is this for? Is it ssn? Then I think we should remove it :dagger:

src/main/kotlin/com/hedvig/underwriter/web/dtos/IncompleteQuoteDto.kt
palmenhq: Idk really, is it? 😅 do you know if product-pricing cares somehow?

FredrikAreschoug: @vonElfvin do you know if product-pricing cares? And by that could we remove it in product-pricing?

PR #23, opened 2019-11-04
Feature/add campaign code to partner

FredrikAreschoug:
~~I haven't added the campaign codes `f89051cb43` and `8a2fcb2a11` which obvious should be done before merging.~~ But let me know what you think. Edit: Created the campaigns in staging and prod

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/model/Partner.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/Partner.kt
@@ -1,7 +1,7 @@
 package com.hedvig.underwriter.model
 
-enum class Partner {
+enum class Partner(val campaignCode: String? = null) {
     HEDVIG,
-    INSPLANET,
-    COMPRICER
+    INSPLANET("f89051cb43"),
+    COMPRICER("8a2fcb2a11")
 }
--- a/src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
@@ -230,7 +230,7 @@ class QuoteServiceImpl(
                     )
                 )
                 if (response.statusCode.isError) {
-                    logger.error("Failed to redeem $campaignCode for parnter ${quote.attributedTo} with status code ${response.statusCodeValue} and reason ${response.statusCode.reasonPhrase} ")
+                    logger.error("Failed to redeem $campaignCode for parnter ${quote.attributedTo} with status code ${response.statusCodeValue} and reason ${response.statusCode.reasonPhrase}")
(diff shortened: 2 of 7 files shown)


src/main/kotlin/com/hedvig/underwriter/model/Partner.kt
```diff
@@ -1,7 +1,7 @@
 package com.hedvig.underwriter.model
 
-enum class Partner {
+enum class Partner(val campaignCode: String? = null) {
     HEDVIG,
-    INSPLANET,
-    COMPRICER
+    INSPLANET("f89051cb43"),
```
meletios: Why not use their names as campaigncodes?  🤔

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -219,6 +221,16 @@ class QuoteServiceImpl(
             val signedQuoteId =
                 productPricingService.createProduct(quoteWithMember.getRapioQuoteRequestDto(body.email), memberId).id
 
+            quote.attributedTo.campaignCode?.let { campaignCode ->
```
meletios: What if we fail to redeem the code ?

src/main/kotlin/com/hedvig/underwriter/model/Partner.kt
meletios: It really depends on how we use the campaign codes

FredrikAreschoug: We did the same with dreams is mostly because no one should guess the code in the on boarding. I know that is highly unlikely even if it's `insplanet` but just to be on the safe side :) . It should only be used here so why not creating something obscure :)

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
FredrikAreschoug: My plan was to just log the error, so we will be notified if it starts happening but but I looked in to product and pricing but we haven't defined what we return if something goes wrong. Can I suspect an `ResponseEntity<*>`

FredrikAreschoug: My plan was to just log the error, so we will be notified if it starts happening but but I looked in to product and pricing but we haven't defined what we return if something goes wrong. Can I suspect an ResponseEntity<*>

meletios: Or `ResponseEntity<Void>` ? But yes, lets log it 👍

PR #24, opened 2019-11-20
Draft functionality probable to change

johantjellden:
----

#

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
@@ -88,7 +88,9 @@ data class Quote(
     val startDate: LocalDate? = null,
 
     val validity: Long = ONE_DAY * 30,
-    val memberId: String? = null
+    val memberId: String? = null,
+
+    val campaign: Campaign? = null
 ) {
     val isComplete: Boolean
         get() = when {
@@ -214,6 +216,10 @@ data class Quote(
         }
 
         return Left(errorStrings)
+    }
+
+    fun addCampaign(campaign: Campaign): Quote {
+        return this.copy(campaign = campaign)
     }
 
     companion object {
--- a/src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/ProductPricingServiceImpl.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/ProductPricingServiceImpl.kt
@@ -1,11 +1,15 @@
 package com.hedvig.underwriter.serviceIntegration.productPricing
 
-import com.hedvig.underwriter.serviceIntegration.productPricing.dtos.ApartmentQuotePriceDto
-import com.hedvig.underwriter.serviceIntegration.productPricing.dtos.HouseQuotePriceDto
-import com.hedvig.underwriter.serviceIntegration.productPricing.dtos.QuotePriceResponseDto
-import com.hedvig.underwriter.serviceIntegration.productPricing.dtos.RapioProductCreatedResponseDto
-import com.hedvig.underwriter.serviceIntegration.productPricing.dtos.RapioQuoteRequestDto
-import com.hedvig.underwriter.serviceIntegration.productPricing.dtos.RedeemCampaignDto
+import arrow.core.Either
+import arrow.core.Left
+import arrow.core.Right
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.hedvig.underwriter.model.*
+import com.hedvig.underwriter.service.QuoteServiceImpl
+import com.hedvig.underwriter.serviceIntegration.productPricing.dtos.*
+import com.hedvig.underwriter.serviceIntegration.productPricing.dtos.Incentive as DtoIncentive
+import feign.FeignException
+import org.slf4j.LoggerFactory
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.cloud.openfeign.EnableFeignClients
 import org.springframework.stereotype.Service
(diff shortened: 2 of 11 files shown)


src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/ProductPricingServiceImpl.kt
```diff
@@ -36,4 +43,24 @@ class ProductPricingServiceImpl @Autowired constructor(
...
+            if(e.status() == 400) {
+                val errorResponse = objectMapper.readTree(e.content())
+                Left(errorResponse["message"].asText())
+            } else {
+                logger.error("Failed to redeem ${redeemCampaignDto.code}  with response ${e.contentUTF8()}")
+                throw e
+            }
+        }
+
+    private fun antiCorruptIncentive(incentive: DtoIncentive): Incentive = when (incentive) {
```
palmenhq: i think we could work on this naming, a) methods should be verbs imo, b) what is an "anti corrupt" incentive?

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/ProductPricingServiceImpl.kt
```diff
@@ -36,4 +43,24 @@ class ProductPricingServiceImpl @Autowired constructor(
...
     override fun redeemCampaign(redeemCampaignDto: RedeemCampaignDto) =
         this.productPricingClient.redeemCampaign(redeemCampaignDto)
+
+    override fun validateRedeemableCampaign(redeemCampaignDto: ValidateCampaignDto): Either<String, Campaign> =
+        try {
+            val body = this.productPricingClient.validateRedeemCampaign(redeemCampaignDto)
+            Right(Campaign( body.body!!.campaignId, antiCorruptIncentive(body.body!!.incentive)))
+        } catch (e:FeignException) {
+            if(e.status() == 400) {
+                val errorResponse = objectMapper.readTree(e.content())
```
palmenhq: what about `readValue(e.content(), AProperErrorDto::class.java)`?

src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
```diff
@@ -216,6 +218,10 @@ data class Quote(
         return Left(errorStrings)
     }
 
+    fun addCampaign(campaign: Campaign): Quote {
```
palmenhq: do we really need this? Isn't `copy` enough? i think this could be misinterpreted as a modifier rather than a "copyer"?

src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
```diff
@@ -88,7 +88,9 @@ data class Quote(
     val startDate: LocalDate? = null,
 
     val validity: Long = ONE_DAY * 30,
-    val memberId: String? = null
+    val memberId: String? = null,
+
+    val campaign: Campaign? = null
```
palmenhq: shall we maybe persist this also?

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/ProductPricingServiceImpl.kt
johantjellden: It is not an "anti corrupt" incentive. It's a function in the [anticorruption layer](http://wiki.c2.com/?AnticorruptionLayer), might be an idea of a pattern taken to far. But I think that the logic for mapping our external incetives into internal ones should live in this package. Adding a method to do the mapping to the DTO also felt weird since we want the DTOs to be quite similar to the ones in the service we are talking to and to keep them low on logic.
So I think that a function at this level is the best choice.

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/ProductPricingServiceImpl.kt
johantjellden: We have not defined what the propper ErrorDTO is, if we do define what is is we can create an exception mapper that can do the mapping for us.

src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
johantjellden: Well this is an immutable class so I don't think that the problem with a modifier is that big, the returntype also signals that it will modify its object.
I think that this is mostly a discussion about where to put logic, when we use copy methods we effectively use our class as a simple struct. I think that this more dangerous, and will lead to us spreading our domain logic around into different methods where we happen to use this class, than collecting it to where it makes "OO" sense to have it.

src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
johantjellden: Yes but I wanted to share the stuff I had written so far, so I would not implement something that needed to be changed.

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/ProductPricingServiceImpl.kt
palmenhq: ahaa, maybe we could just rephrase it as `ensureAntiCorruption` or something could make it clearer? :)

src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
palmenhq: Ah yes, this is true - nice observation. We're mixing a bit currently, hence my confusion. I don't really agree that obfuscating and encapsulating mutations on an object that holds data is good object oriented design but i guess that's totally a preference and up for discussion. My preferred way of doing it would rather be to have an separate, stateless, class (`QuoteService` seems like an excellent candidate) that performs mutations to have clear separation of concerns between who owns the modifications and who holds the data. Right now i believe we spread the logic even more by splitting it between the database model and the service layer.

johantjellden:
Lets close this.

PR #25, opened 2019-11-21
Feature/new web onboarding

FredrikAreschoug:


src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
```diff
@@ -0,0 +1,90 @@
...
+import com.hedvig.underwriter.model.QuoteInitiatedFrom
+import com.hedvig.underwriter.service.QuoteService
+import com.hedvig.underwriter.serviceIntegration.productPricing.ProductPricingService
+import com.hedvig.underwriter.web.dtos.ErrorCodes
+import graphql.schema.DataFetchingEnvironment
+import java.lang.IllegalStateException
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Component
+
+@Suppress("unused")
```
johantjellden: I don't think we need this

src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
```diff
@@ -0,0 +1,9 @@
+package com.hedvig.underwriter.graphql
+
+import com.coxautodev.graphql.tools.GraphQLQueryResolver
+import org.springframework.stereotype.Component
+
+// This is unfortunately needed
+@Suppress("unused")
```
johantjellden: Remove this as well

src/test/kotlin/com/hedvig/underwriter/graphql/GraphQlMutationsTest.kt
```diff
@@ -0,0 +1,91 @@
...
+                houseHoldSize = 2,
+                houseType = ApartmentProductSubType.BRF,
+                isStudent = false
+        ))).thenReturn(
+            QuotePriceResponseDto(
+                BigDecimal.ONE
+            )
+        )
+
+        val response = graphQLTestTemplate.perform("/mutations/createQuote.graphql", null)
```
johantjellden: See if we can inline the mutation into the test somehow.

FredrikAreschoug: Unfortunately, it is not possible to inline it. So either I can stop using these test facilities or keep it as it is. The pros to keep using them is that they have support for variables, that we don't use that much but nice to have.

pom.xml
```diff
@@ -189,6 +200,26 @@
...
         </dependency>
 
+        <!-- GraphQL -->
+        <dependency>
+            <groupId>com.hedvig.graphql.commons</groupId>
+            <artifactId>graphql-commons</artifactId>
+            <version>1.0.4</version>
+        </dependency>
+        <dependency>
+            <groupId>com.graphql-java-kickstart</groupId>
```
meletios: do you need that now?

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
```diff
@@ -0,0 +1,89 @@
...
+@Component
+class Mutation @Autowired constructor(
+    private val quoteService: QuoteService,
+    private val productPricingService: ProductPricingService,
+    private val localizationService: LocalizationService,
+    private val textKeysLocaleResolver: TextKeysLocaleResolver
+) : GraphQLMutationResolver {
+
+    fun createQuote(input: CreateQuoteInput, env: DataFetchingEnvironment): QuoteResult {
+        // TODO: let's discuss initiatedFrom!
```
meletios: Shall we ? 😜

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
```diff
@@ -0,0 +1,89 @@
...
+        val quote = quoteService.createQuote(input.toIncompleteQuoteDto(memberId = env.getTokenOrNull()), input.id, initiatedFrom = QuoteInitiatedFrom.WEBONBOARDING)
+
+        return when (val quoteOrError = quoteService.completeQuote(quote.id)) {
```
meletios: Is it the left side the error? Should we rename `quoteOrError` to `errorOrQuote` ? 😕

meletios: or swap the sides?

pom.xml
FredrikAreschoug: es it's the test facilities, didn't figure out how to add them to the package

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
FredrikAreschoug: o yeah! How do we detect if it's an app or a web client doing this mutation?

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
FredrikAreschoug: true I will switch the name feels like a big change to swap sides

PR #29, opened 2019-11-25
Added a test data builder

johantjellden:
An databuilder usable with quotes


johantjellden:
> Very neat! However, maybe we should do this also for the `QuoteRepositoryTest`?

Yes but I lets do that next time we touch that test.

PR #32, opened 2019-11-27
sign flow fixes

FredrikAreschoug:


src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -235,9 +210,47 @@ class QuoteServiceImpl(
...
+                            ErrorResponseDto(
+                                ErrorCodes.MEMBER_HAS_EXISTING_INSURANCE,
+                                "quote is already signed"
+                            )
+                        )
+                    }
+                }
+            }
+
+            val memberId = quote.memberId ?: memberService.createMember()
```
meletios: the quote.memberId is null here or? since we check it in the if statement

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -255,44 +268,48 @@ class QuoteServiceImpl(
...
-                    quote.memberId,
-                    campaignCode,
-                    LocalDate.now()
+        quoteWithProductId.attributedTo.campaignCode?.let { campaignCode ->
+            try {
+                productPricingService.redeemCampaign(
+                    RedeemCampaignDto(
+                        quoteWithProductId.memberId,
+                        campaignCode,
+                        LocalDate.now()
```
meletios:
```suggestion
                        LocalDate.now(ZoneId.of("Europe/Stockholm"))
```

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -190,34 +191,8 @@ class QuoteServiceImpl(
...
-                    return Either.Left(
-                        ErrorResponseDto(
-                            ErrorCodes.MEMBER_HAS_EXISTING_INSURANCE,
-                            "quote is already signed"
-                        )
-                    )
-                }
-            }
+        if (quote.originatingProductId != null) {
+            throw Exception("")
```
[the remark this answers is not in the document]
FredrikAreschoug: good question maybe we should ask @johantjellden again

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -235,9 +210,47 @@ class QuoteServiceImpl(
...
+                    val memberAlreadySigned = memberService.isSsnAlreadySignedMemberEntity(quote.data.ssn!!)
+                    if (memberAlreadySigned.ssnAlreadySignedMember) {
+                        return Either.Left(
+                            ErrorResponseDto(
+                                ErrorCodes.MEMBER_HAS_EXISTING_INSURANCE,
+                                "quote is already signed"
+                            )
+                        )
+                    }
+                }
```
palmenhq: We have some code duplication here,  shall we maybe do something like

```kt
val memberAlreadySigned = when (quote.data) {
    is PersonPolicyHolder<*> -> memberService.isSsn...
    else -> throw ...
  }

if (memberAlreadySigned.ssnAlreadySignedMember) {
  ...
}
```

PR #34, opened 2019-11-28
added get qoute

FredrikAreschoug:
Needed to add something in the `Query` and instead of adding jibberish, Palmen told me to add `quote` while doing that I noticed that the schema a bit off. So I made some changes to it :smile_cat: 

@hedvigoscar @sampettersson Please look at the schema, the out commented stuff will come but for that, I need to fetch some data from `ProductPricing` and `MemberService`


src/main/resources/schema.graphqls
```diff
@@ -101,25 +101,26 @@ input RemoveCurrentInsurerInput {
...
-union QuoteResult = Quote | UnderwritingLimitsHit
+union CreateQuoteResult = CompleteQuote | UnderwritingLimitsHit
 
-type Quote {
+type CompleteQuote {
     id: ID!
+    currentInsurer: String
     price: MonetaryAmountV2!
-    details: QuoteDetails!
+#    details: QuoteDetails! TODO lets add this later
```
hedvigoscar: ☝️

src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
```diff
@@ -1,8 +1,19 @@
...
 
 // This is unfortunately needed
 @Component
-class Query : GraphQLQueryResolver
+class Query @Autowired constructor(
+    private val quoteService: QuoteService
+) : GraphQLQueryResolver {
+
+    fun quote(id: UUID): QuoteResult {
+        TODO()
```
meletios: To do or not to do

FredrikAreschoug: Not TODO :) we need to have something here otherwise it will not be a valid query. And instead of adding any jibberish like a `_ : Boolean` I've added something we will use in the future

PR #35, opened 2019-12-02
Add datadog agent

johantjellden:


johantjellden:
I need to get this out, merging with admin rights.

PR #37, opened 2019-12-03
edit member on create quote

FredrikAreschoug:


src/main/kotlin/com/hedvig/underwriter/extensions/ModelExtensions.kt
```diff
@@ -214,3 +218,19 @@ private fun getDefaultDisplayName(type: ExtraBuildingType): String = when (type)
...
 }
+
+fun CreateQuoteInput.toEditMemberRequest() =
+    this.apartment?.let { apartment ->
+        toEditMemberRequest(apartment.street, apartment.zipCode)
+    } ?: this.house?.let { house ->
+        toEditMemberRequest(house.street, house.zipCode)
+    } ?: throw IllegalStateException("Trying to create EditMemberRequest without apartment and house!")
+
+private fun CreateQuoteInput.toEditMemberRequest(street: String, zipCode: String) = EditMemberRequest(
```
meletios: Shall we put this in the `EdiMemberRequest` we can have something like `EditMemberRequet.fromCreateQuoteinput` ? Its feels better cause we expect to return a `EditMemberRequest` object from the `EditMemberRequest` class. eller?

PR #43, opened 2019-12-09
Feature/mutation edit quote

FredrikAreschoug:


src/main/kotlin/com/hedvig/underwriter/extensions/QuoteExtensions.kt
```diff
@@ -0,0 +1,180 @@
...
+        ?: throw RuntimeException("No ssn on Quote! $this")
+
+val Quote.apartment
+    get() = (data as? ApartmentData)
+
+val Quote.house
+    get() = (data as? HouseData)
+
+val Quote.validTo
+    get() = this.createdAt.plusMillis(this.validity)!!
```
palmenhq: Shouldn’t this be `plusSeconds` or?

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
```diff
@@ -53,71 +52,81 @@ class Mutation @Autowired constructor(
...
+        this.getContext<GraphQLServletContext?>()?.httpServletRequest?.getHeader("User-Agent")?.contains(
+            "Android",
+            false
+        )
             ?: false
 
     fun DataFetchingEnvironment.isIOS() =
-        this.getContext<GraphQLServletContext?>()?.httpServletRequest?.getHeader("User-Agent")?.contains("iOS", false)
+        this.getContext<GraphQLServletContext?>()?.httpServletRequest?.getHeader("User-Agent")?.contains(
+            "iOS",
```
palmenhq: I’m fairly sure this won’t work unfortunately. We’d have to check for `iPhone|iPad|iPod|` or something like that if I’m not mistaken

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
```diff
@@ -53,71 +52,81 @@ class Mutation @Autowired constructor(
...
-        TODO()
+                if (quote.isComplete) {
+                    quote.getCompleteQuoteResult(env, localizationService, textKeysLocaleResolver)
+                } else {
+                    quote.getIncompleteQuoteResult(env, localizationService, textKeysLocaleResolver)
+                }
+            }
+        }
+
+    fun getQuoteResultFromError(errorResponse: ErrorResponseDto) = when (errorResponse.errorCode) {
```
palmenhq: I’m aware I’m a bit late to the party but does it really make sense to return an error as a data type? Isn’t it better to utilize GraphQL’s built-in error capabilities? Maybe this has been a topic for discussion I just haven’t been included in but I think this behavior is a bit unexpected from a consumer viewpoint

src/main/kotlin/com/hedvig/underwriter/extensions/QuoteExtensions.kt
FredrikAreschoug: It was plusMillis in `QuoteServiceImpl` before but yes I think you are right! Maybe we should be more clear in the variable name about which time unit we should expect :smile:

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
sampettersson: The iOS app has a custom user-agent which includes the word “iOS” so this would work :)

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
FredrikAreschoug: Yes that is true we should start using GraphQL’s built-in error capabilities. I think we just didn't think about it when we drafted the schema. If we want to change this we should really do it now let's bring it up during standup!

sampettersson: Yeah, I agree with you, altough GraphQL errors are untyped and really hard to interpret in the apps, so we opted for a more explicit way to handle errors for things that often result in specific identifiable errors. That was the reasoning if I remember correctly 😄

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
palmenhq: If we're matching specifically the app maybe we should be a bit more specific in that case?  or?

src/main/kotlin/com/hedvig/underwriter/extensions/QuoteExtensions.kt
palmenhq: Hehe, yes that would be a good idea i think!

PR #46, opened 2019-12-10
Fix/get latest

FredrikAreschoug:


src/main/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImpl.kt
```diff
@@ -51,6 +51,16 @@ class QuoteRepositoryImpl(private val jdbi: Jdbi) : QuoteRepository {
         return findQuote(databaseQuote, dao)
     }
 
+    override fun findLatestOneByMemberId(memberId: String): Quote? =
+        jdbi.inTransaction<Quote?, RuntimeException> { h -> findLatestOneByMemberId(memberId, h) }
+
+    fun findLatestOneByMemberId(memberId: String, h: Handle): Quote? {
+        val dao = h.attach<QuoteDao>()
+        val databaseQuotes = dao.findByMemberId(memberId)
+        val latestQuote = databaseQuotes.maxBy { it.timestamp }!!
```
palmenhq: i believe this could be nullable no? i think we should just return null if it is

PR #48, opened 2019-12-13
insurance cost and misc

FredrikAreschoug:


src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
```diff
@@ -1,31 +1,53 @@
...
-            }
-            else -> {
-                quote.getIncompleteQuoteResult(env, localizationService, textKeysLocaleResolver)
-            }
-        }
+        quote.toResult(env)
     } ?: throw IllegalStateException("No quote found!")
+
+    fun lastQuoteOfMember(env: DataFetchingEnvironment) =
+        quoteRepository.findLatestOneByMemberId(env.getToken())?.let { quote ->
```
palmenhq: imo this should go into `quoteService` rather than using the repository directly. Feels more java :)))

src/main/kotlin/com/hedvig/underwriter/graphql/type/QuoteResult.kt
```diff
@@ -8,10 +8,12 @@ sealed class QuoteResult {
...
     data class CompleteQuote(
         val id: UUID,
-        val firstName: String,
-        val lastName: String,
+        val firstName: String?,
+        val lastName: String?,
+        val ssn: String?,
         val currentInsurer: String?,
         val price: MonetaryAmountV2,
+        val insuranceCost: InsuranceCost,
```
palmenhq: do we need `price` now?

src/main/resources/schema.graphqls
```diff
@@ -113,6 +114,10 @@ type CompleteQuote {
     id: ID!
     currentInsurer: String
     price: MonetaryAmountV2!
+    insuranceCost: InsuranceCost!
+    firstName: String
+    lastName: String
+    ssn: String
```
palmenhq: `!`?

src/main/resources/schema.graphqls
```diff
@@ -279,4 +287,17 @@ type IncompleteHouseQuoteDetails {
...
+    numberOfBathrooms: Int
+    yearOfConstruction: Int
+    numberOfBathrooms: Int
+    isSubleted: Boolean
 }
+
+type InsuranceCost {
+    monthlyGross: MonetaryAmountV2!,
+    monthlyDiscount: MonetaryAmountV2!,
+    monthlyNet: MonetaryAmountV2!,
```
palmenhq: do we really always have a discount? or will it just be `0`?

FredrikAreschoug: It will be 0, and changing it could break the stitching in giraffe

src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
```diff
@@ -1,31 +1,53 @@
...
-                quote.getIncompleteQuoteResult(env, localizationService, textKeysLocaleResolver)
-            }
-        }
+        quote.toResult(env)
     } ?: throw IllegalStateException("No quote found!")
+
+    fun lastQuoteOfMember(env: DataFetchingEnvironment) =
+        quoteService.getLatestQuoteForMemberId(env.getToken())?.let { quote ->
+            quote.toResult(env)
+        } ?: throw IllegalStateException("No quote found!")
```
palmenhq: "... for memberId ${env.getToken()}" maybe? otherwise this error gets quite useless imo

palmenhq:
Screenshot 2019-12-16 at 16 28 14

PR #49, opened 2019-12-13
add sentry logging

AlexSJ1990:


src/main/kotlin/com/hedvig/underwriter/serviceIntegration/customerio/CustomerIO.kt
```diff
@@ -1,23 +1,31 @@
...
 @EnableFeignClients
-class CustomerIO(val customerIOClient: CustomerIOClient) {
+class CustomerIO(val customerIOClient: CustomerIOClient, val env: Environment) {
 
     fun setPartnerCode(memberId: String, partner: Partner) {
         try {
             val map = mutableMapOf("partner_code" to partner.name)
             this.customerIOClient.put(memberId, map)
         } catch (ex: Exception) {
+            if (Arrays.stream(env.activeProfiles).anyMatch { env -> env.contentEquals("staging") ||
```
johantjellden: This check needs to happen one level above. Where we use CustomerIO.

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -326,7 +330,15 @@ class QuoteServiceImpl(
         quoteRepository.update(signedQuote, signedAt)
 
         if (quoteWithProductId.attributedTo != Partner.HEDVIG) {
+            try {
             customerIOClient?.setPartnerCode(quoteWithProductId.memberId, quoteWithProductId.attributedTo)
+            } catch (exception: Exception) {
+                if (Arrays.stream(env.activeProfiles).anyMatch { env -> env.contentEquals("staging") ||
```
palmenhq: pretty sure `env.activeProfiles.any { env -> env.contentEquals("staging") }` would be sufficient :)
https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/any.html

PR #51, opened 2019-12-16
Fix/missing input and responses

FredrikAreschoug:


src/main/kotlin/com/hedvig/underwriter/graphql/type/CurrentInsurer.kt
```diff
@@ -0,0 +1,67 @@
...
+import java.lang.IllegalArgumentException
+
+data class CurrentInsurer(
+    val id: String,
+    val displayName: String,
+    val switchable: Boolean
+) {
+    companion object {
+        fun create(id: String): CurrentInsurer {
+            return when (id) {
```
palmenhq: i feel like this could be done a whole lot prettier, what about making a `Map` of `id` -> `Pair<Name, WhateverTheBooleanMeans>`? and then just something like

```kt
val theInsurer = theMap[id] ?: throw ...
return CurrentInsurer(id, theinsurer.first, theInsurer.second)
```
or so?

src/main/kotlin/com/hedvig/underwriter/util/DateUtil.kt
```diff
@@ -0,0 +1,14 @@
+package com.hedvig.underwriter.util
+
+import java.time.Instant
+import java.time.LocalDateTime
+import java.time.ZoneId
+
+private val STOCKHOLM_ZONE_ID: ZoneId = ZoneId.of("Europe/Stockholm")!!
+
+fun Instant.toLocalDate() =
+    LocalDateTime.ofInstant(this, STOCKHOLM_ZONE_ID).toLocalDate()
```
palmenhq: does `LocalDate` not have an `ofInstant`? i think i've seen different approaches but this one seems unfamiliar

PR #59, opened 2020-01-08
added email and data collection id to quote

FredrikAreschoug:


src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
```diff
@@ -99,7 +101,9 @@ data class Quote(
     val underwritingGuidelinesBypassedBy: String? = null,
     val memberId: String? = null,
     val originatingProductId: UUID? = null,
-    val signedProductId: UUID? = null
+    val signedProductId: UUID? = null,
+
```
meletios: empty row :iamsorry:

src/main/resources/db/changelog/migrations/20200108-1000-add-email-and-data-collection-id-to-quote.sql
```diff
@@ -0,0 +1,17 @@
+--liquibase formatted.sql
+
+--changeset fredrikareschoug:20200108-1000-add-email-and-data-collection-id-to-quote.sql
+
+ALTER TABLE quote_revision_house_data
+    ADD COLUMN email varchar(100) NULL
```
meletios: why? 100 and not 255? 😇

PR #73, opened 2020-02-23
parse age on norwegian ssn on quote data

FredrikAreschoug:


src/main/kotlin/com/hedvig/underwriter/graphql/ResponseExtensions.kt
```diff
@@ -29,6 +30,7 @@ fun Quote.getCompleteQuoteResult(
     email = email,
     currentInsurer = currentInsurer?.let { CurrentInsurer.create(it) },
     ssn = ssn,
+    birthDate = birthDate,
```
meletios: I am not sure if we will have the age as an integer though. Can you speak with Zak?

FredrikAreschoug: Talked to zak about it! And he was fine with birth date :) If we change our mind let's drop the column and age instead

PR #74, opened 2020-02-27
todo fixes and set birthdate from ssn on rapio request

FredrikAreschoug:


src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
```diff
@@ -394,10 +432,18 @@ data class Quote(
             CompleteQuoteDetails.UnknownQuoteDetails()
         } ?: this.norwegianTravel?.let {
             CompleteQuoteDetails.UnknownQuoteDetails()
-        } ?: throw IllegalStateException("Trying to create QuoteDetails without `swedishApartment`, `swedishHouse` data")
+        }
+        ?: throw IllegalStateException("Trying to create QuoteDetails without `swedishApartment`, `swedishHouse` data")
```
johantjellden: Seems like this can fail on more than swedishApartment, swedishHouse.

PR #86, opened 2020-03-16
prevent signing quotes with various member ids

FredrikAreschoug:


src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
```diff
@@ -47,10 +47,14 @@ class SignServiceImpl(
     val env: Environment
 ) : SignService {
 
-    override fun startSigningQuotes(quoteIds: List<UUID>, ipAddress: String?): StartSignResponse {
+    override fun startSigningQuotes(quoteIds: List<UUID>, memberId: String, ipAddress: String?): StartSignResponse {
 
         val quotes = quoteService.getQuotes(quoteIds)
         quotes.forEach { quote ->
+            quote.memberId?.let { quoteMemberId ->
+                if (memberId != quoteMemberId) return StartSignResponse.FailedToStartSign(VARIOS_MEMBER_ID_ERROR_MESSAGE)
```
meletios: Shall we log it here also?

PR #87, opened 2020-03-18
Feature/bundle quotes

FredrikAreschoug:
@sampettersson @hedvigoscar @palmenhq Please have a look at the schema. One thought I had about `bundledQuotes` should we have checks on this as we have in `signQuotes`? And with that return a union with an error type?

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/graphql/type/TypeMapper.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/graphql/type/TypeMapper.kt
@@ -5,11 +5,7 @@ import com.hedvig.localization.service.LocalizationService
 import com.hedvig.underwriter.graphql.type.depricated.CompleteQuoteDetails
 import com.hedvig.underwriter.model.ExtraBuilding
 import com.hedvig.underwriter.model.ExtraBuildingType
-import com.hedvig.underwriter.model.NorwegianHomeContentsData
-import com.hedvig.underwriter.model.NorwegianTravelData
 import com.hedvig.underwriter.model.Quote
-import com.hedvig.underwriter.model.SwedishApartmentData
-import com.hedvig.underwriter.model.SwedishHouseData
 import com.hedvig.underwriter.model.birthDate
 import com.hedvig.underwriter.model.email
 import com.hedvig.underwriter.model.firstName
--- a/src/main/kotlin/com/hedvig/underwriter/graphql/type/TypeMapper.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/graphql/type/TypeMapper.kt
@@ -44,12 +44,7 @@ class TypeMapper(
             birthDate = quote.birthDate,
             price = MonetaryAmountV2(
                 quote.price!!.toPlainString(),
-                when (quote.data) {
-                    is SwedishHouseData,
-                    is SwedishApartmentData -> "SEK"
-                    is NorwegianHomeContentsData,
-                    is NorwegianTravelData -> "NOK"
-                }
+                quote.currency
             ),
             quoteDetails = mapToQuoteDetails(quote, locale),
             startDate = quote.startDate,
--- a/src/main/kotlin/com/hedvig/underwriter/graphql/type/TypeMapper.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/graphql/type/TypeMapper.kt
@@ -5,7 +5,11 @@ import com.hedvig.localization.service.LocalizationService
 import com.hedvig.underwriter.graphql.type.depricated.CompleteQuoteDetails
 import com.hedvig.underwriter.model.ExtraBuilding
 import com.hedvig.underwriter.model.ExtraBuildingType
+import com.hedvig.underwriter.model.NorwegianHomeContentsData
+import com.hedvig.underwriter.model.NorwegianTravelData
 import com.hedvig.underwriter.model.Quote
+import com.hedvig.underwriter.model.SwedishApartmentData
+import com.hedvig.underwriter.model.SwedishHouseData
 import com.hedvig.underwriter.model.birthDate
 import com.hedvig.underwriter.model.email
 import com.hedvig.underwriter.model.firstName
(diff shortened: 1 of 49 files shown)


src/main/kotlin/com/hedvig/underwriter/graphql/type/TypeMapper.kt
```diff
@@ -27,6 +31,34 @@ class TypeMapper(
...
+            currentInsurer = quote.currentInsurer?.let { CurrentInsurer.create(it) },
+            ssn = quote.ssn,
+            birthDate = quote.birthDate,
+            price = MonetaryAmountV2(
+                quote.price!!.toPlainString(),
+                when (quote.data) {
+                    is SwedishHouseData,
+                    is SwedishApartmentData -> "SEK"
+                    is NorwegianHomeContentsData,
+                    is NorwegianTravelData -> "NOK"
```
vonElfvin: Feels like this logic could be cetralised somwhere. For example having a quote.data.getCurrency() or quote.data.currency with a 
```
val currency: String
  get() = when (this) {
    is SwedishHouseData,
    is SwedishApartmentData -> "SEK
    is NorwegianHomeContentsData,
    is NorwegianTravelData -> "SEK
  }
```

src/main/kotlin/com/hedvig/underwriter/service/BundleQuotesServiceImpl.kt
```diff
@@ -0,0 +1,58 @@
...
+                        Money.of(
+                            quote.price, when (quote.data) {
+                                is SwedishHouseData,
+                                is SwedishApartmentData -> "SEK"
+                                is NorwegianHomeContentsData,
+                                is NorwegianTravelData -> "NOK"
+                            }
+                        ), when (val data = quote.data) {
+                            is SwedishHouseData -> ProductType.SWEDISH_HOUSE
+                            is SwedishApartmentData -> if (data.isStudent) ProductType.SWEDISH_STUDENT_BRF else ProductType.SWEDISH_BRF
```
vonElfvin: What about rent?

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/dtos/CalculateBundleInsuranceCostRequest.kt
```diff
@@ -0,0 +1,5 @@
+package com.hedvig.underwriter.serviceIntegration.productPricing.dtos
+
+data class CalculateBundleInsuranceCostRequest(
+    val productsToBeBundled: List<CalculateBundledPriceDto>
```
vonElfvin: Either

```suggestion
    val insurancesToBeBundled: List<CalculateBundledPriceDto>
```
or 
```suggestion
    val contractsToBeBundled: List<CalculateBundledPriceDto>
```
Or something, products is kind of a dead concept

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/dtos/ProductType.kt
```diff
@@ -0,0 +1,11 @@
...
+
+enum class ProductType {
+    SWEDISH_BRF,
+    SWEDISH_STUDENT_BRF,
+    SWEDISH_HOUSE,
+    NORWEGIAN_HOME_CONTENTS,
+    NORWEGIAN_TRAVEL,
+    NORWEGIAN_YOUTH_HOME_CONTENTS,
+    NORWEGIAN_YOUTH_TRAVEL
+}
```
vonElfvin: Two things:
- What happened to rent?
- Why the word `Product`, is maybe contract better? Or Insurance?

src/main/resources/schema.graphqls
```diff
@@ -6,14 +6,15 @@ schema {
 type Query {
     quote(id: ID): Quote!
     lastQuoteOfMember: Quote!
+    bundledQuotes(input: BundledQuotesInput!): BundledQuotes!
```
hedvigoscar: How come we're using `Bundled Quotes` instead of `Quotes`?

Is it because a bundle may have a price smaller than the sum of its parts? 🤔

src/main/resources/schema.graphqls
```diff
@@ -474,4 +475,28 @@ type FailedToStartSign {
...
+    price: MonetaryAmountV2!
+    firstName: String!
+    lastName: String!
+    ssn: String!
+    birthDate: LocalDate!
+    quoteDetails: QuoteDetails!
+    startDate: LocalDate
+    expiresAt: LocalDate!
+    email: String
+    dataCollectionId: ID
```
hedvigoscar: What's a `dataCollectionId`?

src/main/resources/schema.graphqls
FredrikAreschoug: In norway you will be able to get a bundle discount and to be able to apply that we will need to know what you are bundling

src/main/resources/schema.graphqls
FredrikAreschoug: it's the insurely data collection id

src/main/resources/schema.graphqls
hedvigoscar: I see 🙂

src/main/resources/schema.graphqls
hedvigoscar: I don't know what that is, nor how it relates to a quote 🤔

FredrikAreschoug: we currently have it on the quote to be able to show the user what him/her pays to there current insurer

hedvigoscar: Makes sense!

hedvigoscar:
I'm ok with the schema ✅

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/dtos/CalculateBundledPriceDto.kt
```diff
@@ -0,0 +1,8 @@
+package com.hedvig.underwriter.serviceIntegration.productPricing.dtos
+
+import org.javamoney.moneta.Money
+
+data class CalculateBundledPriceDto(
+    val grossPrice: Money,
```
palmenhq: `MonetaryAmount` or?

src/main/resources/schema.graphqls
```diff
@@ -474,4 +475,28 @@ type FailedToStartSign {
     errorMessage: String
 }
 
+input BundledQuotesInput {
+    ids: [ID!]!
+}
+
+type BundledQuotes {
```
palmenhq: Whats your thoughts on this name? I'm getting a little confused because im expecting it to be a list of bundles, i think flipping it would make it clearer. So `QuotesBundle`? Indicating that its _a_ bundle of quotes

hedvigoscar: Or even `QuoteBundle` as Bundle implies multiple of the type 🤷‍♂

FredrikAreschoug: let's do `QuoteBundle`

src/main/resources/schema.graphqls
```diff
@@ -474,4 +475,28 @@ type FailedToStartSign {
     errorMessage: String
 }
 
+input BundledQuotesInput {
```
palmenhq: Should this name maybe now follow `QuoteBundle`?

PR #88, opened 2020-03-19
Feature/customerio notification service

johantjellden:


src/main/kotlin/com/hedvig/underwriter/serviceIntegration/customerio/CustomerIOClient.kt
```diff
@@ -2,21 +2,23 @@ package com.hedvig.underwriter.serviceIntegration.customerio
...
+@ConditionalOnProperty(value = ["customerio.username"], matchIfMissing = false)
 @FeignClient(
     name = "customer.io.client",
     url = "\${customerio.url:https://track.customer.io/api}",
     configuration = [FeignConfiguration::class]
 )
 interface CustomerIOClient {
     @PutMapping("/v1/customers/{id}")
-    fun put(@PathVariable id: String, @RequestBody data: Any): ResponseEntity<String>
+    fun put(@PathVariable id: String, @RequestBody data: Map<String, Any?>): ResponseEntity<String>
```
FredrikAreschoug: dose any need to be nullable? I'm thinking maybe we shouldn't send it if it's null

johantjellden: Setting values to null is the way you remove them from customer.io.

PR #91, opened 2020-03-23
Use price-engine for norwegian price calculations

vonElfvin:
- Use new price-engine for Norwegian prices
- Remove old price mock
- Fix tests (hopefully, I'm unable to run clean install locally since the embedded database won't work)
- Update localization package version so the CI is run


src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/dtos/contract/NorwegianTravelLineOfBusiness.kt
```diff
@@ -3,4 +3,11 @@ package com.hedvig.underwriter.serviceIntegration.productPricing.dtos.contract
 enum class NorwegianTravelLineOfBusiness {
     REGULAR,
     YOUTH;
+
+    companion object {
+        fun from(isYouth: Boolean): NorwegianTravelLineOfBusiness = when (isYouth) {
+            true -> YOUTH
+            false -> REGULAR
+        }
```
FredrikAreschoug:
```suggestion
        fun from(isYouth: Boolean): NorwegianTravelLineOfBusiness = if (isYouth) YOUTH else REGULAR
```
I think this is cleaner but do as you please :)

PR #94, opened 2020-03-24
Make ssn nullable for QuoteResult and BundledQuote

vonElfvin:


src/main/resources/schema.graphqls
```diff
@@ -462,7 +462,6 @@ type IncompleteHouseQuoteDetails {
     extraBuildings: [ExtraBuilding!]
     numberOfBathrooms: Int
     yearOfConstruction: Int
-    numberOfBathrooms: Int
```
meletios: do we want that?

vonElfvin: For some reason we had 2x numberOfBathrooms 🤷‍♂ Look at line 463 above

PR #106, opened 2020-04-27
create quote from back-office

AlexSJ1990:


src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/ProductPricingClient.kt
```diff
@@ -79,4 +82,9 @@ interface ProductPricingClient {
         @Valid @RequestBody request: CalculateBundleInsuranceCostRequest,
         @PathVariable memberId: String
     ): ResponseEntity<InsuranceCost>
+
+    @GetMapping("/_/agreements/{agreementId}")
+    fun getAgreementFromAgreementId(
```
FredrikAreschoug, line 87: I think just `getAgreement` is fine

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -148,6 +148,36 @@ class QuoteServiceImpl(
         return transformCompleteQuoteReturn(breachedGuidelinesOrQuote, quoteId)
     }
 
+    override fun createQuoteFromBackOffice(
```
FredrikAreschoug, line 151: In my opinion this should be `createQuoteFromAgreement`

src/main/kotlin/com/hedvig/underwriter/web/QuoteController.kt
```diff
@@ -71,6 +72,20 @@ class QuoteController @Autowired constructor(
             ).getOrHandle { it }
     }
 
+    @PostMapping("/createQuoteFromBackOffice")
```
FredrikAreschoug, line 75: If you like the naming of from agreement change the name of this endpoint and the dto as well :)

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
vonElfvin: I agree

PR #109, opened 2020-05-06
add stub for price engine

AlexSJ1990:


src/main/kotlin/com/hedvig/underwriter/serviceIntegration/priceEngine/PriceEngineServiceStub.kt
```diff
@@ -0,0 +1,16 @@
+package com.hedvig.underwriter.serviceIntegration.priceEngine
+
+import com.hedvig.underwriter.serviceIntegration.priceEngine.dtos.PriceQueryRequest
+import com.hedvig.underwriter.serviceIntegration.priceEngine.dtos.PriceQueryResponse
+import org.javamoney.moneta.Money
+import java.util.*
+
+class PriceEngineServiceStub: PriceEngineService {
+    override fun queryNorwegianHomeContentPrice(query: PriceQueryRequest.NorwegianHomeContent): PriceQueryResponse {
+        return PriceQueryResponse(queryId = UUID.randomUUID(), price = Money.of(109, "NOK"))
```
AlexSJ1990, line 10: should we make this something suspiciously large?

FredrikAreschoug: Yes that would be better, but what is the purpose of the stub? Is it for tests? or running it locally? If so shouldn't we be able to include it in builds like that only? and fail the bull if it is relying on this in stagging or prod

johantjellden: If we use it only for tests we can put it in the test part of the source tree.

PR #110, opened 2020-05-07
Feature/create contract from hope

meletios:


src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
```diff
@@ -139,6 +145,10 @@ class Mutation @Autowired constructor(
             throw IllegalStateException("Invalid state [Error Message: ${errorResponse.errorMessage}]")
         ErrorCodes.UNKNOWN_ERROR_CODE ->
             throw IllegalStateException("Unknown error code [Error Message: ${errorResponse.errorMessage}]")
+        ErrorCodes.MEMBER_HAS_NOT_EXISTING_SIGNED_INSURANCE ->
+            throw IllegalStateException("Member has not existing signed insurance [Error Message: ${errorResponse.errorMessage}]")
```
vonElfvin, line 149:
```suggestion
        ErrorCodes.MEMBER_DOES_NOT_HAVE_EXISTING_SIGNED_INSURANCE ->
            throw IllegalStateException("Member does not have existing signed insurance [Error Message: ${errorResponse.errorMessage}]")
```

src/main/kotlin/com/hedvig/underwriter/service/QuoteService.kt
```diff
@@ -58,4 +58,9 @@ interface QuoteService {
     fun getQuotes(quoteIds: List<UUID>): List<Quote>
 
     fun addAgreementFromQuote(request: AddAgreementFromQuoteRequest): Either<ErrorResponseDto, Quote>
+
+    fun createQuoteForNewContractFromHope(
+        quoteRequest: QuoteRequest,
+        underwritingGuidelinesBypassedBy: String?
```
vonElfvin, line 64: Creating the quote should not require any guidelines passing I think?

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -148,6 +147,41 @@ class QuoteServiceImpl(
         return transformCompleteQuoteReturn(breachedGuidelinesOrQuote, quoteId)
     }
 
+    override fun createQuoteForNewContractFromHope(
+        quoteRequest: QuoteRequest,
+        underwritingGuidelinesBypassedBy: String?
+    ): Either<ErrorResponseDto, CompleteQuoteResponseDto> {
+        val quoteId = UUID.randomUUID()
+
+        val updatedQuoteRequest = updateIncompleteQuoteWithMember(quoteRequest)
```
vonElfvin, line 156: Thinking of it, this should probably also be performed at sign in case any info is updated before?

src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
```diff
@@ -263,6 +264,57 @@ class SignServiceImpl(
...
+
+        if (quote.signedProductId != null) {
+            throw RuntimeException("There is a signed product id ${quote.signedProductId} already")
+        }
+
+        if (quote.memberId == null) {
+            Either.Left(
+                ErrorResponseDto(
+                    ErrorCodes.MEMBER_ID_IS_NOT_PROVIDED,
+                    "no member id connected to quote"
```
vonElfvin, line 282: Not sure what the error is used for, but could be nice to have the id in case it gets logged
```suggestion
                    "No member id connected to quote ${quote.id}"
```

src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
```diff
@@ -263,6 +264,57 @@ class SignServiceImpl(
...
+        if (quoteNotSignableErrorDto != null) {
+            return Either.left(quoteNotSignableErrorDto)
+        }
+
+        val memberAlreadySigned = when (quote.data) {
+            is PersonPolicyHolder<*> -> memberService.isSsnAlreadySignedMemberEntity(quote.data.ssn!!)
+            else -> throw RuntimeException("Unsupported quote data class")
+        }
+
+        if (!memberAlreadySigned.ssnAlreadySignedMember) {
```
vonElfvin, line 297: Smart!

src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
```diff
@@ -263,6 +264,57 @@ class SignServiceImpl(
...
+        val updatedQuote = quote.copy(startDate = request.activationDate)
+
+        return Right(
+            signQuoteWithMemberId(
+                updatedQuote,
+                false,
+                SignRequest("", "", ""),
+                (quote.data as PersonPolicyHolder<*>).email!!
+            )
+        )
```
vonElfvin, line 315: Not sure what this does, but maybe we should skip some steps in the signing process here since there should be no bankid involved? Or maybe it should? Maybe this initiates a sign from H.OPE. rather, that would be cool, however it increases the scope so it should be skipped I think?

src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
AlexSJ1990, line 297: So did we agree that you need to be signed in order to create a contract from h.ope?

meletios: I think its the best approach for the first iteration at least. If you (@AlexSJ1990) and @vonElfvin are on board, i will continue with that. To be honest, i didn't contact anyone from IEX 😇

src/main/kotlin/com/hedvig/underwriter/service/QuoteService.kt
johantjellden: Yes maybe, but if we create a quote and offer that to a member, i think w
e are bound by law to sell it if the member wants.

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -148,6 +147,41 @@ class QuoteServiceImpl(
...
+            underwriter.createQuote(
+                updatedQuoteRequest,
+                quoteId,
+                QuoteInitiatedFrom.HOPE,
+                underwritingGuidelinesBypassedBy
+            )
+        val quote = when (breachedGuidelinesOrQuote) {
+            is Either.Left -> breachedGuidelinesOrQuote.a.first
+            is Either.Right -> breachedGuidelinesOrQuote.b
+        }
```
johantjellden, line 168: do we ever use breachedGuidelinesOrQuote.a.second? Or can we simplify the return from `underwriter.createQuote`

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -148,6 +147,41 @@ class QuoteServiceImpl(
...
+        quoteRepository.insert(quote)
+
+        return transformCompleteQuoteReturn(breachedGuidelinesOrQuote, quoteId)
+    }
+
+    private fun updateIncompleteQuoteWithMember(input: QuoteRequest): QuoteRequest {
```
johantjellden, line 174:
```suggestion
    private fun updateQuoteRequestWithMember(input: QuoteRequest): QuoteRequest {
```

src/main/kotlin/com/hedvig/underwriter/service/QuoteService.kt
meletios: This will follow the existing functionality on creating quotes, so i will keep it as it is.

PR #112, opened 2020-05-10
change type to subType and small typo fix

AlexSJ1990:


src/main/kotlin/com/hedvig/underwriter/graphql/type/NorwegianHomeContentsType.kt
```diff
@@ -1,6 +1,6 @@
 package com.hedvig.underwriter.graphql.type
 
 enum class NorwegianHomeContentsType {
-    RENT,
-    OWN
+    OWN,
+    RENT
```
vonElfvin, line 5: Big upgrade

PR #114, opened 2020-05-13
Fix bug with empty input to QuoteBundle query

johantjellden:
Fixes this [bug](https://sentry.io/organizations/hedvig/issues/1646372882/?project=1773000&query=is%3Aunresolved&statsPeriod=14d) which basically happens because we get an empty list to the `quoteBundle` query.
Please look through the solution especially @palmenhq .

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
@@ -94,7 +94,7 @@ interface QuoteDao {
             ORDER BY qr.master_quote_id ASC, qr.id DESC
         """
     )
-    fun find(@BindList("quoteIds") quoteIds: List<UUID>): List<DatabaseQuoteRevision>
+    fun find(@BindList("quoteIds", onEmpty = BindList.EmptyHandling.NULL_STRING) quoteIds: List<UUID>): List<DatabaseQuoteRevision>
 
     @SqlUpdate(
         """
--- a/src/test/kotlin/com/hedvig/underwriter/graphql/BundleQueryIsolationTest.kt
+++ b/src/test/kotlin/com/hedvig/underwriter/graphql/BundleQueryIsolationTest.kt
@@ -1,0 +1,62 @@
+package com.hedvig.underwriter.graphql
+
+import com.hedvig.graphql.commons.extensions.getAcceptLanguage
+import com.hedvig.graphql.commons.extensions.getToken
+import com.hedvig.localization.service.LocalizationService
+import com.hedvig.localization.service.TextKeysLocaleResolver
+import com.hedvig.localization.service.TextKeysLocaleResolverImpl
+import com.hedvig.underwriter.graphql.type.QuoteBundleInputInput
+import com.hedvig.underwriter.graphql.type.TypeMapper
+import com.hedvig.underwriter.service.BundleQuotesService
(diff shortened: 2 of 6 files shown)


src/test/kotlin/com/hedvig/underwriter/graphql/BundleQueryIsolationTest.kt
```diff
@@ -0,0 +1,62 @@
...
+    fun `need at least one contractId`() {
+
+        val query = Query(
+            quoteService,
+            bundleQuoteService,
+            textKeysLocaleResolver,
+            TypeMapper(localizationService)
+            )
+
+        every { bundleQuoteService.bundleQuotes(any(), any(), any()) } throws RuntimeException("Should not get here")
```
palmenhq, line 55: maybe we could use a relaxed mock and `verify(exactly = 0) { .. }` instead?

src/test/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImplTest.kt
```diff
@@ -554,6 +554,15 @@ class QuoteRepositoryImplTest {
         assertQuotesDeepEqualExceptInternalId(updatedQuote, quoteDao.find(quote.id))
     }
 
+    @Test
+    fun findQuotesWithEmptyList() {
+        val quoteDao = QuoteRepositoryImpl(jdbiRule.jdbi)
+
+        val result = quoteDao.findQuotes(listOf())
+
+        assert(result.isEmpty())
```
palmenhq, line 563: it doesn't really matter but you actually don't need parenthesis for the `assert` keyword

src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
```diff
@@ -94,7 +94,7 @@ interface QuoteDao {
             ORDER BY qr.master_quote_id ASC, qr.id DESC
         """
     )
-    fun find(@BindList("quoteIds") quoteIds: List<UUID>): List<DatabaseQuoteRevision>
+    fun find(@BindList("quoteIds", onEmpty = BindList.EmptyHandling.NULL_STRING) quoteIds: List<UUID>): List<DatabaseQuoteRevision>
```
palmenhq, line 97: do we need this? Considering we fail before it should get here, right?

src/test/kotlin/com/hedvig/underwriter/graphql/BundleQueryIsolationTest.kt
johantjellden: Yes, but then we would need to set up returns for product and pricing as well. I think that this is enough.

src/test/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImplTest.kt
johantjellden: aha, i guess you don't need parenthesis around your single arguments in kotlin?

src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
johantjellden: This is a good question. My thinking is that this is how our dao should work. It feels right for me that if you pass and empty list to a find function you get an empty result back. 
But maybe it will just confuse someone in the future?

PR #115, opened 2020-05-18
Log error as info and remove unnecessary code

johantjellden:


src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -330,13 +330,7 @@ class QuoteServiceImpl(
     }
 
     override fun getQuotes(quoteIds: List<UUID>): List<Quote> {
-        val quotes = quoteRepository.findQuotes(quoteIds)
-
-        return if (quotes.all { it != null }) {
-            quotes.filterNotNull()
-        } else {
-            throw RuntimeException("Could not find all quotes in list: $quoteIds")
-        }
```
FredrikAreschoug, line 339: This doesn't look good! Good that you changed it!

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -330,13 +330,7 @@ class QuoteServiceImpl(
...
+        return quoteRepository.findQuotes(quoteIds)
```
FredrikAreschoug, line 333: But shouldn't we filter not null here? Otherwise, there will be a graphql validate schema error, or do we prefer that if one to them is null?

johantjellden: `quoteRepository.findQuotes(quoteIds)` returns List<Quote> so there can not be any nulls in the list.

johantjellden: And in quote repository we do this:
```
private fun findQuotes(quoteIds: List<UUID>, h: Handle): List<Quote> {
        val dao = h.attach<QuoteDao>()
        val databaseQuote = dao.find(quoteIds)
        return databaseQuote.mapNotNull { findQuote(it, dao) }
    }
```

PR #117, opened 2020-05-18
currentInsurer on AgreementQuote

FredrikAreschoug:
Found a bug where we only send current insurer from the first quote but we need it on quote level


FredrikAreschoug:
Creates a [query](https://redash.prod.hedvigit.com/queries/178/source) for this and I look like we are fine

PR #118, opened 2020-05-22
passing iex token to db

meletios:


src/main/resources/db/changelog/migrations/20200522-1100-add-new-field-to-track-iex-token-on-quote.sql
```diff
@@ -0,0 +1,9 @@
+--liquibase formatted.sql
+
+--changeset meletis:20200522-1100-add-new-field-to-track-iex-token-on-quote.sql
+
+ALTER TABLE quote_revisions
+    ADD COLUMN sign_from_hope_triggered_by varchar(255) NULL
```
vonElfvin, line 6: How long is the token? If it's longer than 255 we have issues here

meletios: i change it to text just to avoid that! 👍

PR #121, opened 2020-06-05
Feature/expose quote by contractid

johantjellden:
For a feature in notification-service I needed access to the attributedTo value for a contract. And I thought that getting some information for a Quote given an contract is a good feature to have. So i created an endpoint for that.

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
@@ -1,13 +1,13 @@
 package com.hedvig.underwriter.model
 
-import java.time.Instant
-import java.util.UUID
 import org.jdbi.v3.sqlobject.customizer.Bind
 import org.jdbi.v3.sqlobject.customizer.BindBean
 import org.jdbi.v3.sqlobject.customizer.BindList
 import org.jdbi.v3.sqlobject.statement.GetGeneratedKeys
 import org.jdbi.v3.sqlobject.statement.SqlQuery
 import org.jdbi.v3.sqlobject.statement.SqlUpdate
+import java.time.Instant
+import java.util.UUID
 
 interface QuoteDao {
     @SqlUpdate(
@@ -322,7 +322,7 @@ interface QuoteDao {
             JOIN quote_revisions qr
             ON qr.master_quote_id = mq.id 
             
-            WHERE qr.contract_id = :contractId
+            WHERE qr.contract_id = :contractId AND qr.originating_product_id IS NULL
             ORDER BY qr.master_quote_id ASC, qr.id DESC
     """
     )
--- a/src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
@@ -31,6 +31,7 @@ interface QuoteDao {
                 underwriting_guidelines_bypassed_by,
                 originating_product_id,
                 signed_product_id,
+                contract_id,
                 data_collection_id,
                 sign_from_hope_triggered_by
             )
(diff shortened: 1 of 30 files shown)


src/test/kotlin/com/hedvig/underwriter/web/QuoteControllerTest.kt
```diff
@@ -124,6 +124,60 @@ internal class QuoteControllerTest {
...
+            createdAt = Instant.now(),
+            productType = ProductType.APARTMENT,
+            state = QuoteState.SIGNED,
+            initiatedFrom = QuoteInitiatedFrom.APP,
+            attributedTo = Partner.COMPRICER,
+            data = SwedishApartmentData(
+                id = UUID.randomUUID(),
+                street = "123 Baker street",
+                city = "Stockholm",
+                // numberOfRooms = 3,
```
FredrikAreschoug, line 143: :dagger:

vonElfvin:
Two thoughts for you to consider briefly:
1. The quotes are "per agreement" basis and not contract basis, just so you are aware, for example you could use "contract.currentAgreement" and ask with that instead
2. If we have moved the member to a new address we risk getting the latest quote present that is attributedTo Hedvig even though they could have come from somewhere else

src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
```diff
@@ -301,4 +306,23 @@ interface QuoteDao {
...
+
+            qr.*, 
+            mq.created_at, 
+            mq.initiated_from 
+            
+            FROM master_quotes mq
+            JOIN quote_revisions qr
+            ON qr.master_quote_id = mq.id 
+            
+            WHERE qr.signed_product_id = :contractId
```
vonElfvin, line 323: This will never hit, we use the agreementId and not the contractId when saying "signed_product_id"

vonElfvin: Also, for this use case we should probably get the FIRST where signed_product_id IS NOT NULL instead. Otherwise you risk taking a quote that is created from hope and attributedTo is Hedvig.

vonElfvin: Just to clarify, there is no data point with the `contractId` present in underwriter at all. Either use memberId or agreementId

vonElfvin: Or, if I'm wrong, something has changed since I created this and we probably have a bug in our system

johantjellden: You are most likely correct. I will change this to be based on the agreement Id instead. What I need to to get the correct attributed to for a specific signed contract.

src/test/kotlin/com/hedvig/underwriter/testhelp/databuilder/a.kt
```diff
@@ -68,7 +69,8 @@ class a {
             underwritingGuidelinesBypassedBy,
             memberId,
             originatingProductId,
-            signedProductId
+            agreementId,
+            contractId = contractId
```
FredrikAreschoug, line 73:
```suggestion
            contractId
```
it's okay but stranged to have just the last on as a named argument :)

src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
```diff
@@ -193,7 +195,7 @@ class SignServiceImpl(
             )
         }
 
-        return SignedQuoteResponseDto(signedContractId, signedAt)
+        return SignedQuoteResponseDto(agreementId, signedAt)
```
FredrikAreschoug, line 198: Just to make this a bit cleaner shouldn't we add `agreementId` and `contractId` to `SignedQuoteResponseDto` and keep id but depreciate that field?

johantjellden: We discussed this and decided on sending contractId since hope and rapio are the only external services that get this response. Hope doesn't persist this and for rapio it makes more sense to return the contractId.

PR #126, opened 2020-07-06
Do not throw exception when no ssn exists when sending quote created event

vonElfvin:


src/main/kotlin/com/hedvig/underwriter/serviceIntegration/notificationService/NotificationServiceImpl.kt
```diff
@@ -9,6 +9,7 @@ import com.hedvig.underwriter.model.email
 import com.hedvig.underwriter.model.firstName
 import com.hedvig.underwriter.model.lastName
 import com.hedvig.underwriter.model.ssn
```
meletios, line 11: Is this needed then? 😅

PR #127, opened 2020-07-07
Fix issue when creating quote from youth home content agreement

vonElfvin:


src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/dtos/Agreement.kt
```diff
@@ -110,7 +115,16 @@ sealed class Agreement {
             livingSpace = this.squareMeters.toInt(),
             city = null,
             coInsured = this.numberCoInsured,
-            subType = NorwegianHomeContentsType.valueOf(this.lineOfBusiness.name),
+            subType = when (this.lineOfBusiness) {
```
akashpatl, line 118: sub type or main type?

vonElfvin: Good question, I guess the "main type" is HomeContent in this case

PR #128, opened 2020-07-10
Check if member has contract instead of a signed member when creating contract from hope

vonElfvin:
This is probably how it should have been from the start regardless


meletios:
Its not building

vonElfvin:
Since this is not a real problem we might as well close

PR #135, opened 2020-09-18
Proof of concept jsonschema based quotes

johantjellden:


src/test/kotlin/com/hedvig/underwriter/web/QuoteSchemaTest.kt
```diff
@@ -0,0 +1,366 @@
...
+
+    @Autowired
+    private lateinit var mockMvc: MockMvc
+
+    @MockkBean
+    lateinit var quoteService: QuoteService
+
+    /**
+     *
+     * Quote does not exits
```
AlexSJ1990, line 30: what does this mean?

src/main/kotlin/com/hedvig/underwriter/web/V2QuoteController.kt
```diff
@@ -0,0 +1,53 @@
...
+@RestController
+@RequestMapping("/_/v2/quotes")
+class V2QuoteController(
+    val quoteService: QuoteService,
+    objectMapper: ObjectMapper
+) {
+
+    val jsonSchemaConfig = JsonSchemaConfig
+        .html5EnabledSchema()
+        .withJsonSchemaDraft(JsonSchemaDraft.DRAFT_07)
```
AlexSJ1990, line 31: Can we name something more explicit than DRAFT_07

AlexSJ1990 (dismissed):
Looks really nice and great tests 👍

src/main/kotlin/com/hedvig/underwriter/web/V2QuoteController.kt
johantjellden: DRAFT_07 is the version of the json schema :)

src/test/kotlin/com/hedvig/underwriter/web/QuoteSchemaTest.kt
johantjellden: This was my test list, will remove

src/main/kotlin/com/hedvig/underwriter/web/V2QuoteController.kt
```diff
@@ -0,0 +1,53 @@
...
+import java.util.UUID
+
+@RestController
+@RequestMapping("/_/v2/quotes")
+class V2QuoteController(
+    val quoteService: QuoteService,
+    objectMapper: ObjectMapper
+) {
+
+    val jsonSchemaConfig = JsonSchemaConfig
```
vonElfvin, line 29: Maybe this should be a bean in some @Configuration class

src/main/kotlin/com/hedvig/underwriter/web/V2QuoteController.kt
```diff
@@ -0,0 +1,53 @@
...
+        .html5EnabledSchema()
+        .withJsonSchemaDraft(JsonSchemaDraft.DRAFT_07)
+    val jsonSchemaGenerator = JsonSchemaGenerator(objectMapper, false, jsonSchemaConfig)
```
vonElfvin, line 32: Same here, a bean makes more sense IMO

src/main/kotlin/com/hedvig/underwriter/web/V2QuoteController.kt
```diff
@@ -0,0 +1,53 @@
...
+
+        val dataClass = when (quote.data) {
+            is SwedishHouseData -> QuoteRequestData.SwedishHouse::class.java
+            is SwedishApartmentData -> QuoteRequestData.SwedishApartment::class.java
+            is NorwegianHomeContentsData -> QuoteRequestData.NorwegianHomeContents::class.java
+            is NorwegianTravelData -> QuoteRequestData.NorwegianTravel::class.java
+        }
+
+        return jsonSchemaGenerator.generateJsonSchema(dataClass)
+    }
```
vonElfvin, line 52: The return type should be a Schema and the data for that schema. Otherwise we cannot prepopulate the form once we have the schema. Both of them JsonNodes or similar

src/main/kotlin/com/hedvig/underwriter/web/V2QuoteController.kt
```diff
@@ -0,0 +1,53 @@
...
+
+    @GetMapping("{quoteId}/schema")
+    fun getSchemaForQuote(@PathVariable quoteId: UUID): Any {
+        val quote =
+            quoteService.getQuote(quoteId) ?: return ResponseEntity.status(404).body(
+                ErrorResponseDto(
+                    ErrorCodes.NO_SUCH_QUOTE,
+                    errorMessage = "QuoteNotFound"
+                )
+            )
```
vonElfvin, line 42: How come we don't use notFound here?

```suggestion
            quoteService.getQuote(quoteId) ?: return ResponseEntity.notFound().body(
                ErrorResponseDto(
                    ErrorCodes.NO_SUCH_QUOTE,
                    errorMessage = "QuoteNotFound"
                )
            )
```

vonElfvin: Also, maybe the errorMessage could be more explicit

src/main/kotlin/com/hedvig/underwriter/web/V2QuoteController.kt
```diff
@@ -0,0 +1,53 @@
...
+
+        val dataClass = when (quote.data) {
+            is SwedishHouseData -> QuoteRequestData.SwedishHouse::class.java
+            is SwedishApartmentData -> QuoteRequestData.SwedishApartment::class.java
+            is NorwegianHomeContentsData -> QuoteRequestData.NorwegianHomeContents::class.java
+            is NorwegianTravelData -> QuoteRequestData.NorwegianTravel::class.java
```
vonElfvin, line 48: I think we'll have to create new classes for the schemas and mappers. Since QuoteRequestData includes a bunch of data we don't care about in H.OPE. and we will want to add validation for example which will make them really cluttered if they do both things. Also more things should be required in H.OPE. than are nullable in the QuoteRequestData.

vonElfvin: Also, how will we solve the mutation part where we get the request back of any of the types but they can look identical.

vonElfvin (commented):
Really neat! Added some things to consider

johantjellden:
https://github.com/HedvigInsurance/underwriter/pull/136 supersedes this one.

PR #136, opened 2020-09-28
Feature/json schema update

vonElfvin:
TODO: Fix tests

Schema:
```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "street": {
      "type": "string",
      "title": "Street"
    },
    "zipCode": {
      "type": "string",
      "title": "Zip Code",
      "minLength": 5,
      "maxLength": 5
    },
    "city": {
      "type": "string",
      "title": "City"
    },
    "livingSpace": {
      "type": "integer",
      "title": "Living Space",
      "minimum": 0
    },
    "numberCoInsured": {
      "type": "integer",
      "title": "Number Co-Insured",
      "minimum": 0
    },
    "ancillaryArea": {
      "type": "integer",
      "title": "Ancillary Area",
      "minimum": 0
    },
    "yearOfConstruction": {
      "type": "integer",
      "title": "Year Of Construction",
      "minimum": 0
    },
    "numberOfBathrooms": {
      "type": "integer",
      "title": "Number Of Bathrooms",
      "minimum": 0
    },
    "isSubleted": {
      "type": "boolean",
      "title": "Is Subleted"
    },
    "extraBuildings": {
      "title": "Extra Buildings",
      "type": "array",
      "items": {
        "type": "object",
        "properties": {
          "type": {
            "type": "string",
            "enum": [
              "GARAGE",
              "CARPORT",
              "SHED",
              "STOREHOUSE",
              "FRIGGEBOD",
              "ATTEFALL",
              "OUTHOUSE",
              "GUESTHOUSE",
              "GAZEBO",
              "GREENHOUSE",
              "SAUNA",
              "BARN",
              "BOATHOUSE",
              "OTHER"
            ],
            "title": "Type",
            "default": "GARAGE"
          },
          "area": {
            "type": "integer",
            "title": "Area",
            "minimum": 0
          },
          "hasWaterConnected": {
            "type": "boolean",
            "title": "Has Water Connected"
          }
        },
        "required": [
          "type",
          "area",
          "hasWaterConnected"
        ],
        "title": "Extra Buildings"
      }
    }
  },
  "required": [
    "street",
    "zipCode",
    "livingSpace",
    "numberCoInsured",
    "ancillaryArea",
    "yearOfConstruction",
    "numberOfBathrooms",
    "isSubleted",
    "extraBuildings"
  ],
  "$id": "SwedishHouse"
}
```


src/main/kotlin/com/hedvig/underwriter/service/QuoteSchemaServiceImpl.kt
```diff
@@ -0,0 +1,37 @@
...
+import org.springframework.stereotype.Service
+import java.util.UUID
+
+@Service
+class QuoteSchemaServiceImpl(
+    private val quoteService: QuoteService,
+    private val quoteMapper: QuoteMapper,
+    private val schemaGenerator: SchemaGenerator
+) : QuoteSchemaService {
+    override fun getSchemaByQuoteId(quoteId: UUID): JsonNode? {
```
johantjellden, line 20: There is one thing to think bout here. With this implementation we get a one to one mapping between quote and schema. Which in a way makes sence but we also talked a bit about the possebility to have different schemas for a quote that creates a new contract and a quote that modifies a contract.
Maybe this should be get schema for update.

src/main/kotlin/com/hedvig/underwriter/service/QuoteSchemaServiceImpl.kt
```diff
@@ -0,0 +1,49 @@
...
+
+        return when (quote.data) {
+            is SwedishApartmentData -> getSchemaForContract("SWEDISH_APARTMENT")
+            is SwedishHouseData -> getSchemaForContract("SWEDISH_HOUSE")
+            is NorwegianHomeContentsData -> getSchemaForContract("NORWEGIAN_HOME_CONTENT")
+            is NorwegianTravelData -> getSchemaForContract("NORWEGIAN_TRAVEL")
+        }
+    }
+
+    override fun getSchemaForContract(contractType: String): JsonNode? {
```
vonElfvin, line 31: Would really like for `contractType: String` to be an enum but couldn't come up with a good name for it

src/main/kotlin/com/hedvig/underwriter/service/QuoteSchemaServiceImpl.kt
vonElfvin: I think I'll have to disagree here, since a quote is of a certain type and thus has a certain schema. If you want to change a type of `agreement` then you should use a new quote and not a different schema.

FredrikAreschoug (approved):
Looks good to me even though is more places to create dto's when adding a new quote type. We should really look into code generation, this project is getting absurd

PR #138, opened 2020-10-08
return marketInfo from quote

meletios:
Returns the market info based on the quote type. 

This will be used in payment-service as a fail-safe if payment-service cannot determine the market from the contacts


src/main/kotlin/com/hedvig/underwriter/service/QuoteService.kt
```diff
@@ -58,4 +59,5 @@ interface QuoteService {
 
     fun expireQuote(id: UUID): Quote?
     fun getQuoteByContractId(contractId: UUID): Quote?
+    fun getMarketInfoFromLatestQuote(memberId: String): MarketInfo
```
FredrikAreschoug, line 62: I really don't like using latest quote like this but to be fair it will probably be sufficient.

FredrikAreschoug (approved):
Just a question is it not better to go on picked local from member service? When it comes to context this is really out of place. And also cant this be branched of `feature/enable-danish-signing-through-uw` hence I've been trying to merge that branch for 2 days

vonElfvin (commented):
I agree that we should consider the pickedLocale. But IMO the hierarchy is like this

Contract > Quote > Picked Locale

Where the previous is more accurate market than the next.

src/main/kotlin/com/hedvig/underwriter/model/MarketInfo.kt
```diff
@@ -0,0 +1,11 @@
+package com.hedvig.underwriter.model
+
+class MarketInfo(
```
vonElfvin, line 3:
```suggestion
class QuoteMarketInfo(
```

FredrikAreschoug (changes requested):
Just came up with a solution to my comment lets not use the latest quote to decide the market information but use all member's quotes. If they are all, let's say, Norwegian then return the market information otherwise blow up or something because the member is then in a super weird state. I know it can feel like an overkill but I really want to remove the concept of the latest quote

FredrikAreschoug (approved):
:tada:

FredrikAreschoug (approved):

PR #147, opened 2020-10-20
Add street and insurance type to quote created event

vonElfvin:
This will:
- Add additional data to quote created event
- No longer crash quote creation if notification-service is down/fails


src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -148,7 +148,11 @@ class QuoteServiceImpl(
         }
 
         if (quote.memberId != null && quote.email != null) {
-            notificationService.sendQuoteCreatedEvent(quote)
+            try {
+                notificationService.sendQuoteCreatedEvent(quote)
+            } catch (exception: Exception) {
+                logger.error("Unable to send quote created event (quoteId=${quote.id})", exception)
```
AlexSJ1990, line 154: maybe useful to send memberId also?

AlexSJ1990 (approved):

PR #150, opened 2020-10-28
Add isStudent and type to Danish home content

vonElfvin:


src/main/resources/db/changelog/migrations/20201028-1015-support-student-and-type-denmark.sql
```diff
@@ -0,0 +1,11 @@
+--liquibase formatted.sql
+
+--changeset elvingranat:20201028-1015-support-student-and-type-denmark.sql
+
+ALTER TABLE quote_revision_danish_home_content
+    ADD COLUMN is_student boolean NOT NULL DEFAULT false;
+
+ALTER TABLE quote_revision_danish_home_content
+    ADD COLUMN type VARCHAR (20) NOT NULL DEFAULT 'RENT';
```
vonElfvin, line 9: I'd say either default both or neither or?

src/main/resources/schema.graphqls
```diff
@@ -231,7 +238,10 @@ enum TypeOfContract {
     NO_HOME_CONTENT_YOUTH_RENT
     NO_TRAVEL
     NO_TRAVEL_YOUTH
-    DK_HOME_CONTENT
```
vonElfvin, line 234: TODO: Fix content-service and giraffe

src/main/kotlin/com/hedvig/underwriter/service/model/QuoteSchema.kt
```diff
@@ -52,7 +53,7 @@ sealed class QuoteSchema {
         val numberOfBathrooms: Int,
         @get:JsonProperty("isSubleted")
         @param:JsonProperty("isSubleted")
-        @JsonSchema(title = "Is Subleted", required = true)
```
vonElfvin, line 55: "required = true" caused a lot of confusion in H.OPE. so short team we'll just remove it but fix it in hope later

FredrikAreschoug: :sweat_smile:

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/dtos/contract/AgreementQuote.kt
```diff
@@ -85,6 +85,7 @@ sealed class AgreementQuote {
         val lineOfBusiness: NorwegianTravelLineOfBusiness
     ) : AgreementQuote()
 
+    // TODO: Use product-pricing-objects lib instead so I don't have to change this code here
```
FredrikAreschoug, line 88: :+1: should be fairly easy to do, but yeah let's do it in a new pr

src/main/resources/db/changelog/migrations/20201028-1015-support-student-and-type-denmark.sql
FredrikAreschoug: In my opinion we should remove the default and add `UPDATE quote_revision_danish_home_content SET is_student = false, type = 'rent'`

FredrikAreschoug (commented):
Nice :partying_face: :balloon:!

src/main/resources/db/changelog/migrations/20201028-1015-support-student-and-type-denmark.sql
vonElfvin: I'll just update it manually and remove the defaults since it's staging only

FredrikAreschoug (approved):
:balloon:

src/main/resources/db/changelog/migrations/20201028-1015-support-student-and-type-denmark.sql
johantjellden: Don't we need the default since we are adding a NOT NULL column to an existing table?

johantjellden: And one other point. When we have a DEFAULT value we can run both the old and the new versions at the same time which we do with most of our services and is something we probably want keep taking advantage of in the future.

vonElfvin: What do you suggest we do then @johantjellden and @FredrikAreschoug? I honestly don't know what's best

FredrikAreschoug: Yeah! johan is correct! he needed to explain it to me but yeah we should have defaults not nulls

PR #154, opened 2020-11-16
Mask ssn in log.error

vonElfvin:


src/main/kotlin/com/hedvig/underwriter/util/TextUtil.kt
```diff
@@ -0,0 +1,9 @@
+package com.hedvig.underwriter.util
+
+import com.hedvig.underwriter.model.Market
+
+fun String.maskSsn(market: Market) = when (market) {
+    Market.SWEDEN -> this.replaceRange(this.length - 4, this.length, "XXXX")
+    Market.NORWAY -> this.replaceRange(this.length - 5, this.length, "XXXXX")
+    Market.DENMARK -> this.replaceRange(this.length - 4, this.length, "XXXX")
```
meletios, line 8: Denmark and sweden is the same, shall we have an else branch instead? :D It may be useful if we will add more countries

vonElfvin: I'd rather have a complete enum coverage, so if you add a new market you'll have to considerd the masking of the ssn. I could add them on the same line however!

meletios (approved):

PR #159, opened 2020-11-23
Feature/contract strategy

johantjellden:
I made an attempt to break out contract based logic and separate it from logic that is generic for all quotes. The refactoring is not all done but I think that it is far enough to receive feedback and discuss if this is a good way forward.


src/main/kotlin/com/hedvig/underwriter/service/quoteStrategies/SwedishApartmentDataStrategy.kt
```diff
@@ -0,0 +1,40 @@
...
+import com.hedvig.underwriter.serviceIntegration.notificationService.quoteCreatedEvent
+import com.hedvig.underwriter.serviceIntegration.productPricing.ProductPricingService
+import org.javamoney.moneta.Money
+
+class SwedishApartmentDataStrategy(
+    private val debtChecker: DebtChecker,
+    val productPricingService: ProductPricingService
+) : QuoteStrategy() {
+    override fun createNotificationEvent(quote: Quote): QuoteCreatedEvent {
+        require(quote.data is SwedishApartmentData)
```
FredrikAreschoug, line 21: Don't know what I think about this fail-safe. Either we have it in each function of the `Strategy` classes or we remove it and have test to make sure that the correct strategy. Or come to think of it why not have generics?

FredrikAreschoug (commented):
I really like this strategy pattern! A thought I had is should we have `SwedishDataStrategy` as and abstract class and then `SwedishHouseDataStrategy` inherit  `SwedishDataStrategy` so we don't repeat our selves :)

src/main/kotlin/com/hedvig/underwriter/service/quoteStrategies/SwedishApartmentDataStrategy.kt
vonElfvin: Along with Fredrik's comment. If we are going to potentially throw an excepton like this then we should have something along the lines of `createNoficationEventOrThrow` naming so you know of the risk. Or `throws` so you are ready to catch it

src/main/kotlin/com/hedvig/underwriter/service/quoteStrategies/SwedishApartmentDataStrategy.kt
```diff
@@ -0,0 +1,40 @@
...
+        return quoteCreatedEvent(quote, quote.data.street, quote.data.zipCode, quote.data.subType!!.toString())
+    }
+
+    override fun getInsuranceCost(quote: Quote): InsuranceCost {
+        return productPricingService.calculateInsuranceCost(
+            Money.of(quote.price, "SEK"), quote.memberId!!
```
vonElfvin, line 27: Maybe store the currency somewhere.

src/main/kotlin/com/hedvig/underwriter/service/quoteStrategies/DanishHomeContentsDataStrategy.kt
```diff
@@ -0,0 +1,44 @@
...
+
+        return quoteCreatedEvent(
+            quote,
+            quote.data.street,
+            quote.data.zipCode,
+            if (quote.data.isStudent) {
+                "STUDENT"
+            } else {
+                "REGULAR"
+            }
```
vonElfvin, line 24: I know this is the same as it was before. But this is not correctly implemented in the first place. It should be different for rent/brf here. Look at `NorwegianHomeContentsData`

src/main/kotlin/com/hedvig/underwriter/service/quoteStrategies/StrategyService.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.hedvig.underwriter.model.SwedishApartmentData
+import com.hedvig.underwriter.model.SwedishHouseData
+import com.hedvig.underwriter.service.DebtChecker
+import com.hedvig.underwriter.service.guidelines.BaseGuideline
+import com.hedvig.underwriter.serviceIntegration.notificationService.dtos.QuoteCreatedEvent
+import com.hedvig.underwriter.serviceIntegration.productPricing.ProductPricingService
+import org.springframework.stereotype.Service
+
+@Service
+class StrategyService(
```
vonElfvin, line 20: Maybe?
```suggestion
class QuoteStrategyService(
```

vonElfvin (commented):
Interesting!

src/main/kotlin/com/hedvig/underwriter/service/quoteStrategies/DanishHomeContentsDataStrategy.kt
johantjellden: I will create a bug for this so that we can fix it in a separate issue. I don't want to introduce to many changes in this refactoring

src/main/kotlin/com/hedvig/underwriter/service/quoteStrategies/SwedishApartmentDataStrategy.kt
johantjellden: Would make sense but lets do that at another time.

src/main/kotlin/com/hedvig/underwriter/service/quoteStrategies/SwedishApartmentDataStrategy.kt
johantjellden: In this case we are already checking that quoteData is of the correct type in the outer function. So this require is more of a type check. If we ever throw an exception here i would like underwriter to stop processing the request because that would mean that something big is wrong.

PR #163, opened 2020-11-24
Feature/refactor quoteservice updatequote

johantjellden:
I refactored some messy logic in QuoteServiceImpl to make more usage of Either. I think that the outcome is a bit cleaner.

It is more functional thou but I think that once you get used to that it is quite nice.

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/model/QuoteRepository.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/QuoteRepository.kt
@@ -1,6 +1,5 @@
 package com.hedvig.underwriter.model
 
-import arrow.core.Either
 import java.time.Instant
 import java.util.UUID
 
@@ -12,7 +11,6 @@ interface QuoteRepository {
     fun findOneByMemberId(memberId: String): Quote?
     fun findLatestOneByMemberId(memberId: String): Quote?
     fun expireQuote(id: UUID): Quote?
-    fun <A> modify(quoteId: UUID, modifier: (Quote?) -> Either<A, Quote>): Either<A, Quote>
     fun update(updatedQuote: Quote, timestamp: Instant = Instant.now()): Quote
     fun findByContractId(contractId: UUID): Quote?
 }
--- a/src/main/kotlin/com/hedvig/underwriter/model/QuoteRepository.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/QuoteRepository.kt
@@ -1,5 +1,6 @@
 package com.hedvig.underwriter.model
 
+import arrow.core.Either
 import java.time.Instant
 import java.util.UUID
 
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/hedvig/underwriter/model/QuoteRepository.kt
```diff
@@ -11,7 +12,7 @@ interface QuoteRepository {
     fun findOneByMemberId(memberId: String): Quote?
     fun findLatestOneByMemberId(memberId: String): Quote?
     fun expireQuote(id: UUID): Quote?
-    fun modify(quoteId: UUID, modifier: (Quote?) -> Quote?): Quote?
+    fun <A> modify(quoteId: UUID, modifier: (Quote?) -> Either<A, Quote>): Either<A, Quote>
```
FredrikAreschoug, line 15: Maybe we should discuss this, but I don't see why we need to impose `Either` on the repository. In my head, this is something QuoteService should take care of.

johantjellden: Oh yes I can see it now.

johantjellden: No i was confused for a while. The problem is the code below:
```
quoteRepository.modify(quote.id) { quoteToUpdate ->
                val updatedQuote = quoteToUpdate!!.update(quoteRequest)
                underwriter.validateAndCompleteQuote(updatedQuote, underwritingGuidelinesBypassedBy)
                    .bimap(
                        { errors ->
                            throw QuoteCompletionFailedException(
                                "Unable to complete quote: ${errors.second}",
                                errors.second
                            )
                        },
                        { quote -> quote }
                    ).orNull()
            }!!.let { updatedQuote -> Either.right(updatedQuote) }
```

Since we are running functions that return `Either` inside of the modifier function in the repository it is better if we can return them directly instead of throwing and catching errors.

But maybe the questions is if we need the modify function at all or if we instead should use 
`fun update(updatedQuote: Quote, timestamp: Instant = Instant.now()): Quote`.

I will take a look and see how we use them. The modify function seems a bit wierd.

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -45,68 +46,54 @@ class QuoteServiceImpl(
...
+            // .flatMap { q ->
+            //     if (q.state != QuoteState.QUOTED && q.state != QuoteState.INCOMPLETE) {
+            //         Either.Left(
+            //             ErrorResponseDto(
+            //                 ErrorCodes.INVALID_STATE,
+            //                 "quote [Id: ${q.id}] must be quoted to update but was really ${q.state} [Quote: $q]"
+            //             )
+            //         )
+            //     } else Either.Right(q)
+            // }
```
FredrikAreschoug, line 69: :dagger:

FredrikAreschoug (commented):
Nice yes this is a good improvement :slightly_smiling_face:

FredrikAreschoug (approved):
To me this looks like a nice improvement but remove that blob of commented code. We will find it in the git history if we need it

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
johantjellden: Will do!

PR #168, opened 2020-12-02
store breached uwgl codes in db

FredrikAreschoug:
I did this since it was so easy to do what do you think should we re-add this?


AlexSJ1990 (approved):
I think this is nice to store 👍

vonElfvin (approved):

PR #169, opened 2020-12-08
HVG-538 Metrics for breached guidelines

FredrikAreschoug:
This is to have metrics for breached guidelines in data dog.

Changes:
* created metrics component
* increments on `validateGuidelines`

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/service/Metrics.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/service/Metrics.kt
@@ -1,68 +1,33 @@
 package com.hedvig.underwriter.service
 
 import com.hedvig.underwriter.model.Market
-import com.hedvig.underwriter.service.guidelines.NorwegianHomeContentsGuidelines
-import com.hedvig.underwriter.service.guidelines.NorwegianPersonGuidelines
-import com.hedvig.underwriter.service.guidelines.NorwegianTravelGuidelines
-import com.hedvig.underwriter.service.guidelines.SwedishApartmentGuidelines
-import com.hedvig.underwriter.service.guidelines.SwedishHouseGuidelines
-import com.hedvig.underwriter.service.guidelines.SwedishPersonalGuidelines
 import io.micrometer.core.instrument.Counter
 import io.micrometer.core.instrument.MeterRegistry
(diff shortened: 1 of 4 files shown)


src/main/kotlin/com/hedvig/underwriter/service/Metrics.kt
```diff
@@ -0,0 +1,68 @@
...
+import io.micrometer.core.instrument.Counter
+import io.micrometer.core.instrument.MeterRegistry
+import org.slf4j.LoggerFactory
+import org.springframework.stereotype.Component
+
+@Component
+class Metrics(private val registry: MeterRegistry, depthChecker: DebtChecker) {
+
+    private val breachedUnderwritingGuidelineCounters: Map<Market, Map<String, Counter>>
+
```
johantjellden, line 19: What do you think of another approach where we create the counters the first time they are used?
So perhaps we create a counter map for each country and the we create the counter the first time we increment it.

Then we would not need to touch this file as we add new guidelines and the implementation would be simpler.

The implication it would have is that guidelines that are never breached would not show up in our dashboards.

FredrikAreschoug: Yeah that sounds good

src/main/kotlin/com/hedvig/underwriter/service/Metrics.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.hedvig.underwriter.model.Market
+import io.micrometer.core.instrument.Counter
+import io.micrometer.core.instrument.MeterRegistry
+import org.springframework.stereotype.Component
+
+@Component
+class Metrics(private val registry: MeterRegistry) {
+
+    private val breachedUnderwritingGuidelineCounters = mutableMapOf<Market, MutableMap<String, Counter>>()
```
johantjellden, line 11: So the question is if we should make the maps mutable and have some kind of concurrency protection around breachedUnderwritingGuidelineCounters.

johantjellden: Or maybe use a ConcurrentHashMap

PR #172, opened 2020-12-11
HVG-647 creating sign strategy

meletios:


src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/QuotesExtensions.kt
```diff
@@ -0,0 +1,26 @@
...
+import com.hedvig.underwriter.model.ssn
+import com.hedvig.underwriter.service.model.PersonPolicyHolder
+
+fun List<Quote>.getSSN(): String {
+
+    val ssn = this.first { it.data is PersonPolicyHolder<*> }.ssn
+
+    if (this.any { (it.data as PersonPolicyHolder<*>?)?.let { it.ssn != ssn } == true }) {
+        throw RuntimeException("ssn is not matching when getting ssn from quotes")
+    }
```
johantjellden, line 13:
```
    val ssnSet = this
        .mapNotNull {
            when (it.data) {
                is PersonPolicyHolder<*> -> it.data.ssn
                else -> null
            }
        }
        .toSet()

    if (ssnSet.isEmpty()) {
        throw RuntimeException("list of quotes does not contain any ssns")
    }

    if (ssnSet.size > 1) {
        throw RuntimeException("ssn is not matching when getting ssn from quotes")
    }

    return ssnSet.first()
```

src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/SignStrategyService.kt
```diff
@@ -0,0 +1,65 @@
...
+                when (quotes[0].data) {
+                    is SwedishApartmentData,
+                    is SwedishHouseData -> swedishBankIdSignStrategy.startSign(quotes, signData)
+                    is NorwegianHomeContentsData,
+                    is NorwegianTravelData,
+                    is DanishHomeContentsData -> redirectSignStrategy.startSign(quotes, signData)
+                    is DanishAccidentData,
+                    is DanishTravelData -> StartSignResponse.FailedToStartSign("", "")
+                }
+            2 -> {
```
johantjellden, line 31: Perhaps another way to do this would be to extract out the strategy given all the quotes. Then the cases below could be handled in each strategy?

johantjellden: I guess what I dont like is to have this logic in the middle of this function. Another way could be to assert in the beginning that all quotes are valid. And the select the correct strategy.

johantjellden (commented):
👍 Had some comments

src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/QuotesExtensions.kt
johantjellden: I don't know if this is better :)

FredrikAreschoug (dismissed):
Wow

PR #177, opened 2020-12-15
tests for sign strategy service

FredrikAreschoug:
Here comes the missing tests and a small fix


src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/SignStrategyService.kt
```diff
@@ -31,35 +31,49 @@ class SignStrategyService(
...
                         }
                     }
+                    else -> null
                 }
             }
+            else -> StartSignErrors.quotesCanNotBeBundled
+        }
+
+        if (bundleErrorOrNull != null) {
+            return bundleErrorOrNull
```
meletios, line 76: Isnt it much more clear as it was before?

meletios (approved):
Sweet one minor comment

src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/SignStrategyService.kt
FredrikAreschoug: yes but the test `start sign norwegian and danish quote returns can not be bundled` fails then :slightly_frowning_face:

PR #180, opened 2021-01-07
add and use a neat logger extension to reduce clutter for regular logger creation

bystam:


FredrikAreschoug (approved):
:tada:

vonElfvin (approved):
To production 🚀

PR #182, opened 2021-01-08
Store sign method in sign session

FredrikAreschoug:
In we need to know the sign method for noway.

Changes:
* Store sign method
* Fallbacks to old way of determining sign method

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/model/SignSessionDao.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/SignSessionDao.kt
@@ -12,7 +12,7 @@ interface SignSessionDao {
     @SqlUpdate(
         """
             INSERT INTO sign_sessions (
-                id, signMethod, created_at
+                id, sign_method, created_at
             )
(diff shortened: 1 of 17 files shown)


src/main/kotlin/com/hedvig/underwriter/model/SignSessionDao.kt
```diff
@@ -11,14 +12,14 @@ interface SignSessionDao {
     @SqlUpdate(
         """
             INSERT INTO sign_sessions (
-                id, created_at
+                id, signMethod, created_at
```
bystam: Mixed camel and snake case?

FredrikAreschoug: nice catch :slightly_smiling_face:

FredrikAreschoug:
Looking at this some more this is not the best sulotion

FredrikAreschoug:
Looking at this some more this is probably not the best solution :slightly_smiling_face:

PR #183, opened 2021-01-11
Removed enable simple sign header

FredrikAreschoug:
We are done testing this let's remove it but it is still only enabled in staging


vonElfvin (approved):
👍

PR #185, opened 2021-01-18
Fix/finalize onboarding

vonElfvin:
# Jira Issue: [HVG-737] 

## What?
- Validate that all quotes have the same personal information when attempting to sign
- Call `finalizeOnboarding` when a sign has started

## Why?
- To ensure the latest member personal information is present on the member at the end of onboarding, currently offer page edits of personal info are not committed to the member in member-service

## Optional checklist
- [x] Boyscouted
- [x] Unit tests written
- [ ] Tested locally

[HVG-737]: https://hedvig.atlassian.net/browse/HVG-737

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
@@ -37,7 +37,6 @@ import com.hedvig.underwriter.web.dtos.SignRequest
 import com.hedvig.underwriter.web.dtos.SignedQuoteResponseDto
 import com.hedvig.underwriter.web.dtos.UnderwriterQuoteSignRequest
 import feign.FeignException
-import org.springframework.core.env.Environment
 import org.springframework.stereotype.Service
 import java.time.Instant
 import java.time.LocalDate
@@ -52,8 +51,7 @@ class SignServiceImpl(
     val productPricingService: ProductPricingService,
     val signSessionRepository: SignSessionRepository,
     val signStrategyService: SignStrategyService,
-    val customerIO: CustomerIO,
-    val env: Environment
+    val customerIO: CustomerIO
 ) : SignService {
 
     override fun startSigningQuotes(
@@ -86,13 +84,7 @@ class SignServiceImpl(
             }
         }
 
-        val personalInfoMatching = quotes.windowed(2).all { (predecessor, successor) ->
-            predecessor.firstName == successor.firstName &&
-                predecessor.lastName == successor.lastName &&
-                predecessor.ssn == successor.ssn &&
-                predecessor.email == successor.email &&
-                predecessor.birthDate == successor.birthDate
-        }
+        val personalInfoMatching = bundlePersonalInfoMatching(quotes)
         if (!personalInfoMatching) {
             logger.error("Member [id: $memberId] tried to sign bundle with mismatching personal info [Quotes: $quotes]")
             return StartSignErrors.personalInfoNotMatching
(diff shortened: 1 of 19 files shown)


src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
```diff
@@ -65,26 +70,42 @@ class SignServiceImpl(
...
 
             val quoteNotSignableErrorDto = assertQuoteIsNotSignedOrExpired(quote)
             if (quoteNotSignableErrorDto != null) {
+                logger.error("Member [id: $memberId] tried to sign quote that was either signed or expired [Quotes: $quotes]")
                 return StartSignErrors.fromErrorResponse(quoteNotSignableErrorDto)
             }
         }
 
-        return signStrategyService.startSign(
+        val personalInfoMatching = quotes.windowed(2).all { (predecessor, successor) ->
```
meletios, line 89: Not sure if the words "predecessor, successor" is the correct choice here, since we are just checking if the info is matching between quotes and the quotes are not succeeding one another. (My mother tongue is Greek, so i may be completely wrong). Plus we can extract it to a private function.
```suggestion
        val personalInfoMatching = quotes.windowed(2).all { (left, right) ->
```

meletios (dismissed):
Cool, one small comment though

src/test/kotlin/com/hedvig/underwriter/service/SignServiceImplTest.kt
```diff
@@ -211,6 +213,10 @@ class SignServiceImplTest {
 
         verify(exactly = 1) { signSessionRepository.insert(any()) }
         assertThat(result).isInstanceOf(StartSignResponse.SwedishBankIdSession::class.java)
+
+        verify {
+            memberService.finalizeOnboarding(quote, quote.email!!)
+        }
```
johantjellden, line 219: These tests are getting a bit hairy perhaps we should look at splitting this and the tests below into smaller pieces.

src/test/kotlin/com/hedvig/underwriter/service/SignServiceImplTest.kt
```diff
@@ -246,6 +252,9 @@ class SignServiceImplTest {
 
         assertThat(result).isInstanceOf(StartSignResponse.FailedToStartSign::class.java)
         assertThat((result as StartSignResponse.FailedToStartSign).errorMessage).isEqualTo("Failed")
+        verify(inverse = true) {
+            memberService.finalizeOnboarding(any(), any())
+        }
```
johantjellden, line 257: Perhaps we don't need to check for the inverse case? It would simplify the test and the consequence of an extra `finalizeOnboarding` is not that problematic or?

johantjellden (commented):
Nice work! Extra points for awesome testing 🥇

src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
vonElfvin: I think the naming is decent based on the function name "windowed" that passes through the list, thus one of the quotes will be the "predecessor" and the other the "successor" of each other. The "successor" will be the "predecessor" when it moves one more step. But "left" and "right" could be sufficient if you imagine it moving from left to right

src/test/kotlin/com/hedvig/underwriter/service/SignServiceImplTest.kt
vonElfvin: I think it's good to ensure it doesn't call it since it could be either triggered from hope, or fail to start a sign. Both of these scenarios should explicitly not call the endpoint and are thus tested for :)

PR #188, opened 2021-01-21
Allow sign if ssn is not on quote itself

vonElfvin:
## What?
- Backfill SSN if present on member

## Why?
- Members in Norway who create bundle through webonboarding cannot activate it later since there is no ssn present on the quote

## Optional checklist
- [ ] Boyscouted
- [ ] Unit tests written
- [ ] Tested locally

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
@@ -14,6 +14,7 @@ import com.hedvig.underwriter.model.email
 import com.hedvig.underwriter.model.firstName
 import com.hedvig.underwriter.model.lastName
 import com.hedvig.underwriter.model.ssn
+import com.hedvig.underwriter.model.ssnMaybe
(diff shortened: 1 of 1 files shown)


AlexSJ1990 (approved):

FredrikAreschoug (commented):
I think this only will break down the line or? Will we not need to fetch the ssn the strategies as well?

johantjellden:
I don't really understand all of this, why do we have the ssn on member-service but not on the quote? And I think we at least want a test the new logic so that we can understand why it is there. :)

johantjellden:
Hmm. 
1: If you are an existing member in norway and have and existing insurance. You can then go to the web onboarding and only get the other insurance type?
2: When you sign up for that other insurance we don't know your ssn until when you _sign_ with zignsec?

How do you end up in this state? Don't every web-onboarding session get their own memberId? In wich case we cannot use the memberId to lookup your ssn.

vonElfvin:
This issue havn't happened in a while. May as well close and wait for the problem to arise again

PR #190, opened 2021-01-26
Make undewriter not require an application-development.yml at all

bystam:
# Jira Issue: [HVG-823] 

## What?
Merge values from both production and development config into the regular `application.yml`, making sure secrets are read as environment variables instead.

This is a first step that makes underwriter no longer need an `application-development.yml`. It is however **not** enough to load secrets from any external store. This solution is just a first step to verify that merging the config files works.

## Why?
This makes `underwriter` runnable with only the dev-env setup guide.

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

[HVG-823]: https://hedvig.atlassian.net/browse/HVG-823


src/main/resources/application.yml
```diff
@@ -25,20 +30,38 @@ graphiql:
...
 ---
 spring:
-  profiles: staging
+  profiles: production, staging
+
+server.port: 80
+logging.level.com.hedvig: TRACE
+
+cloud.aws:
+  region.static: eu-central-1
```
bystam, line 51: This was `eu-west-1` for staging. Does it have to be different for production and staging?

src/main/resources/application.yml
```diff
@@ -1,12 +1,17 @@
 server.port: 5698
+management.endpoints.web.exposure.include: '*'
```
bystam, line 2: This was only configured in production and staging, but it looked like just a harmless value that could/should be applicable to the entire application

src/main/resources/application.yml
```diff
@@ -1,12 +1,17 @@
...
     allow-bean-definition-overriding: true
   datasource:
+    platform: POSTGRESQL
     driverClassName: org.postgresql.Driver
-    url: jdbc:postgresql://localhost:5432/underwriter
-    username: postgres
+    url: ${POSTGRES_URL}/underwriter
+    username: ${POSTGRES_USERNAME}
+    password: ${POSTGRES_PASSWORD}
+    initialization-mode: never
```
bystam, line 14: Same actual configuration for all datasources, and let it differ based on the URL and credentials.

src/main/resources/application.yml
```diff
@@ -25,20 +30,38 @@ graphiql:
...
-graphcms.url: https://api-euwest.graphcms.com/v1/cjmawd9hw036a01cuzmjhplka/master
+graphcms:
+  project: Underwriter
+  url: https://api-euwest.graphcms.com/v1/cjmawd9hw036a01cuzmjhplka/master
+
+lokalise:
+  useFakes: true
+
+hedvig:
+  notification-service.url: http://notification-service
```
bystam, line 41: This looks like it's basically a constant. Maybe it could/should just be inlined in code?

src/main/kotlin/com/hedvig/underwriter/localization/LocalizationConfiguration.kt
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.context.annotation.Configuration
+import java.util.Locale
+
+@Configuration
+class LocalizationConfiguration {
+
+    @Bean
+    @ConditionalOnProperty(
+        value = ["lokalise.projectId", "lokalise.apiToken"]
+    )
```
bystam, line 18: This thing (that I've seen being used elsewhere too) made me able to remove one config value (`lokalise.useFakes` ) from the config. I think keeping the config compact can be neat, to avoid "not seeing the forest for the trees"-scenarios.

What do you think?

src/main/resources/application.yml
```diff
@@ -25,20 +30,31 @@ graphiql:
...
+  stack.auto: false
+
+lokalise:
+  projectId: ${LOKALISE_PROJECT_ID}
+  apiToken: ${LOKALISE_API_TOKEN}
 
 graphql:
-  servlet:
+  servlet: # Should this be true only for staging or production too?
     exception-handlers-enabled: true
```
FredrikAreschoug, line 54: This should only be true in staging, this is so that we can get better graphql errors in staging. This is for the webbis and appies to understand what went wrong. Why we don't want that in prod is because then we don't get any sentry alerts when this is applied. I tried to fix this by ordering but did not manage to :slightly_frowning_face:

FredrikAreschoug (commented):
Nice work! I like it, but I think you should act on the comment about `graphql.servlet.exception-handlers-enabled`

src/main/resources/application.yml
bystam: Ah okay! :) I'll make sure to place it only under staging then. Thanks

bystam: Updated!

src/main/kotlin/com/hedvig/underwriter/localization/LocalizationConfiguration.kt
```diff
@@ -0,0 +1,33 @@
...
+    ): LocalizationService {
+        val client = LokaliseClient(projectId, apiToken)
+        return object : LocalizationService {
+            override fun getTranslation(key: String, locale: Locale): String? = client.getTranslation(key, locale)
+        }
+    }
+
+    @Bean
+    @ConditionalOnMissingBean
+    fun fakeService(): LocalizationService = object : LocalizationService {
```
johantjellden, line 30: Nice 🎉
Should we have a warning here so that we don't accidentally start with the fake one in production?
Perhaps check if the profile is production?

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/notificationService/NotificationServiceClient.kt
```diff
@@ -9,7 +9,7 @@ import org.springframework.web.bind.annotation.RequestBody
 
 @FeignClient(
     name = "notification-service",
-    url = "\${hedvig.notification-service.url:http://notification-service}"
+    url = "http://notification-service"
 )
```
johantjellden, line 13: The reason we configured these was so that we could set them to alternative values when developing. So in my application.development I would set it to `localhost:4087` or something. 
I don't know if we still need that functionality or if there is a better way?

src/main/resources/application.yml
johantjellden: 👍

src/main/resources/application.yml
johantjellden: It was probably a miss configuration. Some services were configured to think they were in eu-west-1 because we needed to talk to some services that were only available there. (SES,..?) This is not the case with underwriter so it makes sense to set it to eu-central-1 for all profiles.

src/main/resources/application.yml
```diff
@@ -40,5 +61,4 @@ graphql:
 spring:
   profiles: test, development
 
-lokalise:
-  useFakes: true
+  liquibase.enabled: true
```
johantjellden, line 64: I would leave this off and instead use the command line parameters, this is defined in `com.hedvig.underwriter.config.LiquibaseRunner` 
The reason i would do that is that when we enable liquibase this way migrations are always run, which is good. But running rollbacks becomes more troublesome.

src/main/resources/application.yml
johantjellden: The reason we configured these was so that we could set them to alternative values when developing. So in my application.development I would set it to localhost:4087 or something.
I don't know if we still need that functionality or if there is a better way?

johantjellden (commented):
Nice work! Left some comments but I dont know if we need to act on any of them :)

src/main/resources/application.yml
bystam: And "moving" it does not affect anything? Data and such is sharded and accessible from all data centers?

src/main/kotlin/com/hedvig/underwriter/localization/LocalizationConfiguration.kt
bystam: Good idea :) will add it.

src/main/resources/application.yml
bystam: Is there a way to check in those parameters? Because it would be a shame if every developer needs to add it manually to their run configurations.

src/main/resources/application.yml
bystam: Right - I see. Then I think we should check in `localhost:4087` for development and set it differently for production. I think `application.yml` makes perfect sense for this, I just didn't have any `application-development.yml` entry of my own to notice it by :) will restore it

bystam: Also - how are requests routed in the kubernetes kluster btw? I first though that we used the same kind of special ports for all the different applications, but it looks like we're using port 80 in production? And looking at this config value, are we using different hostnames for the different services?

src/main/resources/application.yml
bystam: Note that this is only set for development and test.

src/main/resources/application.yml
johantjellden: Yes. The only effect is that we change which endpoint the AWS sdk uses. So when it is set to`eu-west-1` all calls goes to that region. This was handy in the beginning when we did not know how we would create our own aws clients.
By default it looks this up using the AWS metadata service that all ec2 instances has access to.

But I think that when running locally during development there is some reason as to why it is good to set it to a static region.

src/main/resources/application.yml
johantjellden: Yes in kubernetes all services get their own dns name so we can have them all running on port 80. So the custom ports are only needed when we run locally.

src/main/resources/application.yml
johantjellden: Aha yes for development and test I guess we can have `liquibase.enabled: true`.

johantjellden (approved):
Sweet! Left some comments about the aws region thing. 
Lets merge this thing to master and 🎸

src/main/resources/application.yml
bystam: It was like that in the `application-development.yml` I was sent by someone :)

PR #193, opened 2021-02-23
use correct currency in quoteMapper

FredrikAreschoug:
## What?
- while testing rapio we found this :gem: 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


bystam (approved):
Looks simple enough 😅

FredrikAreschoug:
Will be fix in #194

PR #194, opened 2021-02-25
Add native price currency support in Underwriter

mightyost:
# Jira Issue: [HVG-965] 

## What?
Added support for having price currency in quote. We get the price and currency from `PriceEngine` but before this the currency was not used and it was instead hardcoded depending on the insurance type. With this commit the currency from `PriceEngine` is used and also preserved on the quote in database.

Note: This commit adds a new column `currency` on the `quote_revisions` table and back populates it depending on the insurance type. 

This PR also adds the insurance market to the sign request. This is needed by Rapio to be able to resolve the correct payment url to send to signing member.

## Why?
Currencies should not be hardcoded when we have it correctly from the `PriceEngine`.

To support multi currency support we should aim to always have a currency defined when there is a price amount as was not the case with quotes in the database. 

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [x] Tested locally

[HVG-965]: https://hedvig.atlassian.net/browse/HVG-965


src/main/kotlin/com/hedvig/underwriter/model/QuoteData.kt
```diff
@@ -32,6 +32,14 @@ sealed class QuoteData {
...
         }
     }
+
+    fun market(): Market {
+        return when (this) {
+            is SwedishHouseData, is SwedishApartmentData -> Market.SWEDEN
+            is DanishAccidentData, is DanishTravelData, is DanishHomeContentsData -> Market.DENMARK
+            is NorwegianTravelData, is NorwegianHomeContentsData -> Market.NORWAY
+        }
+    }
```
bystam, line 42: This can be a `val` to highlight that it's a simple deduction and not significant work.

```suggestion
    val market: Market get() = when (this) {
        is SwedishHouseData, is SwedishApartmentData -> Market.SWEDEN
        is DanishAccidentData, is DanishTravelData, is DanishHomeContentsData -> Market.DENMARK
        is NorwegianTravelData, is NorwegianHomeContentsData -> Market.NORWAY
    }
}
```

src/main/resources/db/changelog/migrations/20210224-1526-add-currency-field-to-quote-revisions.sql
```diff
@@ -0,0 +1,17 @@
...
+    ADD COLUMN currency varchar(3) NULL
+;
+
+UPDATE quote_revisions set currency = 'NOK' where quote_norwegian_home_contents_data_id is not null;
+UPDATE quote_revisions set currency = 'NOK' where quote_norwegian_travel_data_id is not null;
+UPDATE quote_revisions set currency = 'DKK' where quote_danish_home_contents_data_id is not null;
+UPDATE quote_revisions set currency = 'DKK' where quote_danish_accident_data_id is not null;
+UPDATE quote_revisions set currency = 'DKK' where quote_danish_travel_data_id is not null;
+UPDATE quote_revisions set currency = 'SEK' where quote_apartment_data_id is not null;
+UPDATE quote_revisions set currency = 'SEK' where quote_house_data_id is not null;
```
bystam, line 15: Should this migration end with making `currency` non-null on the table maybe?

bystam (commented):
Love it! I usually find storing/snapshotting relevant data like makes for simpler and more portable solutions.

Just had one question and another stylistic suggestion.

src/main/kotlin/com/hedvig/underwriter/model/QuoteData.kt
mightyost: Thanks for the feedback. Will update to this.

I really appreciates feedback like this. My Kotlin skills is still very immature and this helps me to progress me learning it faster. <3!

src/main/resources/db/changelog/migrations/20210224-1526-add-currency-field-to-quote-revisions.sql
mightyost: From what I understood talking with @FredrikAreschoug is that this should be nullable in db since there can exist incomplete quotes without a price (which is also nullable in the table)...

mightyost:
I have another question for you while I am at the currency thingy in Underwriter...

Today is only `price` returned in these quote api endpoints via the `CompleteQuoteResponseDto`:
  - createQuoteForNewContract
  - createQuote
  - createQuoteForMemberBySchemaData
  - createQuoteFromAgreement

Personally I would like to add the currency to those responses (it would not take long for me to fix). Do you agree?

src/main/resources/db/changelog/migrations/20210224-1526-add-currency-field-to-quote-revisions.sql
vonElfvin: This is true, although I wish it wasn't :( Price should not be nullable, along with other things.

FredrikAreschoug: The ideal solution is to put the price as a nullable object that contains a nonnullable amount and currency

FredrikAreschoug (commented):
I think this looks good, and will be a good solution. Sorry for not mention it yesterday if we would want to make it crisper we could have price as a nullable object that contains a non-nullable amount and currency

PR #196, opened 2021-02-26
add a public version of the 'create quote' endpoint, with a slightly tighter API

bystam:
# Jira Issue: [HVG-859] 

## What?
Adds a public controller for creating quotes, making it accessible to Giraffe.

## Why?
A specific Giraffe mutation would like to mix-and-match som existing endpoints in services, which is rather difficult (albeit possible) in GraphQL. To make it simpler, we simply expose a public version of the existing "create quote" to be used in the moving flow project.

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally

[HVG-859]: https://hedvig.atlassian.net/browse/HVG-859


FredrikAreschoug (approved):
Looks good!

PR #197, opened 2021-03-02
Added support for Rapio to get quote bundle costs.

mightyost:
# Jira Issue: [IPC-34, IPC-40] 

## What?
Added new endpoints in `QuoteController` to support creating quote bundles and signing such.

Changes:
- Added new endpoints in `QuoteController`:
```
    @PostMapping("/bundle")
    fun quoteBundle(@RequestBody @Valid request: QuoteBundleRequestDto): QuoteBundleResponseDto

    @PostMapping("/bundle/signFromRapio")
    fun signQuotesFromRapio(
        @RequestBody request: SignQuotesRequestDto
    ): ResponseEntity<Any>
```

- Added `ErrorException` class to be able to bail out with an API response code without using `Either`s

- Replaced checks on insurance type to use market instead. This makes code more generic and decreases the number of places where to add new logic when introducing new types:
  - Member nationality is resolved from quote market
  - Sign strategies uses quote market instead of insurance type

- Added bundle combo validation in `BundleQuoteService` (which is done in sign too)

- Broke out bundle validation in `SignStrategyService` and uses it to validate bundles in other flows:
  - Sign bundle endpoint
  - Create quote bundle endpoint

## Why?
Needed in Rapio for Norwegian partners.

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [x] Tested locally


src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
```diff
@@ -39,11 +39,17 @@ class Query @Autowired constructor(
...
-            LocaleResolver.resolveLocale(env.getAcceptLanguage())
+            input.ids
         )
+
+        return QuoteBundle(bundle.quotes.map {
+            quoteMapper.mapToBundleQuote(
+                it,
+                LocaleResolver.resolveLocale(env.getAcceptLanguage())
+            )
+        }, bundle.cost)
```
Michael-Duivestein, line 52:
```suggestion
        return QuoteBundle(
                bundle.quotes.map {
                    quoteMapper.mapToBundleQuote(
                        it,
                        LocaleResolver.resolveLocale(env.getAcceptLanguage())
                    )
                }, 
                bundle.cost
        )
```

src/main/kotlin/com/hedvig/underwriter/web/dtos/QuoteBundleResponseDto.kt
```diff
@@ -0,0 +1,38 @@
...
+)
+
+data class Amount(
+    val amount: BigDecimal,
+    val currency: String
+) {
+    companion object {
+        fun of(amount: MonetaryAmountV2) = Amount(BigDecimal(amount.amount), amount.currency)
+    }
+}
```
bystam, line 38: Since these two are supporting types of `QuoteBundleResponseDto`, maybe they should be nested inside of it rather than top level in the package?

bystam (approved):
Left a comment, but it's purely aesthetic so I will approve anyway :)

src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
mightyost: Hmm, so I follow u. It just looks like its the formatting you are pinpointing, right?

I ran this through the `mvn antrun:run@ktlint` and it did not complain at least...

I should maybe test `mvn antrun:run@ktlint-format` and see if it aligns with your suggestion...

Michael-Duivestein: Yes, just the formatting. It's a little hard to easily read what's going on.
Also, the formatters are a bit schizophrenic. I've seen them format the same block of code in three different places in three different ways.
As a result, it's a good idea to QC the auto-formatter.

src/main/kotlin/com/hedvig/underwriter/web/dtos/QuoteBundleResponseDto.kt
mightyost: For sure, will be fixed!

src/main/kotlin/com/hedvig/underwriter/service/BundleQuotesService.kt
```diff
@@ -1,9 +1,14 @@
 package com.hedvig.underwriter.service
 
-import com.hedvig.underwriter.graphql.type.QuoteBundle
-import java.util.Locale
+import com.hedvig.underwriter.graphql.type.InsuranceCost
```
FredrikAreschoug, line 3: Good that you stopped the `Quote` from leaking into the service, but shouldn't we do the same for `InsuranceCost`? :slightly_smiling_face:

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -280,7 +280,7 @@ class QuoteServiceImpl(
 
         val quoteNotSignableErrorDto = assertQuoteIsNotSignedOrExpired(quote)
         if (quoteNotSignableErrorDto != null) {
-            Either.left(quoteNotSignableErrorDto)
+            return Either.left(quoteNotSignableErrorDto)
```
FredrikAreschoug, line 283: Good catch!

FredrikAreschoug (approved):
Wow amazing work! Added a small comment if you feel you have to change a lot to fix it please ignore it! Better to get this out so we can test it!

PR #198, opened 2021-03-04
Fix for always returning currency even if price not sat.

mightyost:
# Jira Issue: [] 

## What?
Fix for always returning currency even if price not sat. Enforcing this in earlier commit broke backwards compatibility in `getLatestQuoteFromMemberId` and `getAllQuotesFromMemberId` in quote api.

## Why?
Hope is broken

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/dtos/QuoteDto.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/dtos/QuoteDto.kt
@@ -55,7 +55,7 @@ data class QuoteDto(
                 id = quote.id,
                 createdAt = quote.createdAt,
                 price = quote.price,
-                currency = quote.currency,
+                currency = quote.currencyWithFallbackOnMarket,
                 productType = quote.productType,
                 state = quote.state,
                 initiatedFrom = quote.initiatedFrom,


bystam (approved):

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/dtos/QuoteDto.kt
```diff
@@ -55,7 +55,7 @@ data class QuoteDto(
                 id = quote.id,
                 createdAt = quote.createdAt,
                 price = quote.price,
-                currency = quote.currency,
+                currency = quote.currencyWithFallbackOnMarket,
```
vonElfvin, line 58: This feels 1% more correct 
```suggestion
                currency = quote.currency ?: quote.currencyWithFallbackOnMarket,
```

Michael-Duivestein (approved):

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/productPricing/dtos/QuoteDto.kt
mightyost: This is made implicit in the `quote.currencyWithFallbackOnMarket` getter...

vonElfvin: Ah then it's perfect

PR #199, opened 2021-03-04
rewrite the Dockerfile to support build+test+assembling, with integration tests

bystam:
# Jira Issue: [HVG-1041] 

## What?
Give the dockerfile the power to build both integration test and production images.

## Why?
This is our way to achieve very similar builds for all services, letting the differences be spelled out in the repo itself using Docker.
https://docs.google.com/spreadsheets/d/1PVW1vsiT7fccsCsqzbOalOYEDZAoJotzhsNMho9p1VE/edit#gid=0

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally

[HVG-1041]: https://hedvig.atlassian.net/browse/HVG-1041


Dockerfile
```diff
@@ -1,6 +1,52 @@
...
+FROM build AS integration_test
+COPY src/test src/test
+RUN mvn test-compile -s /usr/share/maven/ref/settings-docker.xml
+
+RUN yum -y install python3 \
+    python3-pip \
+    shadow-utils \
+    util-linux
+RUN adduser underwriter
+USER underwriter
```
bystam, line 36: These 3 commands came from `./testRunnerImage/Dockerfile`

pom.xml
```diff
@@ -482,6 +482,7 @@
                     <execution>
                         <goals>
                             <!-- Sets git config specified under configuration > gitConfig. -->
+                            <goal>initialize</goal>
```
bystam, line 485: Our maven builds fail if there is no git repo because of this plugin. Adding this makes it create an empty git repo instead of failing. This is needed since we don't copy the git files in the dockerfile.

palmenhq (approved):
lgtm!

PR #200, opened 2021-03-05
Logging incoming request data.

mightyost:
# Jira Issue: [] 

## What?
This PR logs out the request data in `QuoteController`, `Mutation` and `Query`.

The request PII fields are masked out.

There are also some temporary logging in `SignServiceImpl` that is needed for me to understand how it is used in order to be able to refactor it without breaking anything. These loggings will be removed later. 

**A word about the PII masking**
Sensitive fields are annotated with the `@Pii` annotation. Calling `.toNonPiiString()` on any object will generate same result as calling `.toString()` but with the `@Pii` annotated field's values sat to `***`.

## Why?
Make it easier to troubleshoot errors, support cases and understand behaviour of the service.

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [x] Tested locally


bystam (commented):
Veeery cool!

I have a follow-up thing I wonder about this. Is there any way we can easily start logging state in a structured form instead of just stringified in the message? I know some of these logger libraries has a way to attach structured data.

I am thinking being able to do something like this:
```kotlin
logger.info(
    "Creating quote",
    data = mapOf("quoteData", quote)
)
```
and have said data attached to the message in a structured manner. Having structured data enables much more advanced querying and aggregating.

I believe doing `MDC.put(...)` would achieve the same thing but that requires one to clear it all the time too, which is a bit annoying.

src/main/kotlin/com/hedvig/underwriter/util/Pii.kt
```diff
@@ -0,0 +1,39 @@
...
+
+            val pii = prop.getAnnotation(Pii::class.java)
+            val value = if (pii != null) "***" else prop.get(obj)?.toNonPiiString()?.trim()
+
+            s += "${prop.name}=" + value
+        }
+        clazz = clazz.superclass
+    }
+    return "${obj.javaClass.simpleName}(${s.joinToString(", ")})"
+}
```
bystam, line 39: Maybe performance isn't a problem of ours... but do you see any risk that reading all of this runtime metadata every time could make things slow. And is there a way to create a cache for some of these lookups?

Thinking of something like a:
```kotlin
// the Boolean is whether the field is @Pii or not
private val cache: ConcurrentHashMap<Class<*>, Pair<Field, Boolean>> = ...
```
That would at least avoid *some* of the reflection reads every time. Maybe it won't make that much of a difference though... part of me thinks that `Field.get(Object o)` is the most expensive operation, and that one is hard to cache.

bystam: Also maybe I should be quiet about performance in our "very low traffic" universe :D

src/main/kotlin/com/hedvig/underwriter/util/Pii.kt
```diff
@@ -0,0 +1,39 @@
...
+
+import java.lang.reflect.Modifier
+import java.util.LinkedList
+
+@Target(AnnotationTarget.FIELD)
+@Retention(AnnotationRetention.RUNTIME)
+annotation class Pii()
+
+fun Any?.toNonPiiString(): String =
+    if (this == null) "null" else reflectionToString(this)
```
bystam, line 11: Should this one maybe check for its conformance to `Map` and `Collection` first?

```suggestion
fun Any?.toNonPiiString(): String = when (this) {
  null -> "null"
  is Map<*, *> -> this.toNonPiiString()
  is Collection<*> -> this.toNonPiiString()
  else -> reflectionToString(this)
```

bystam: I think otherwise that which implementation is run is decided at compile time, since the two implementations below aren't actually overrides.

src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
```diff
@@ -25,16 +27,27 @@ class Query @Autowired constructor(
...
+        logger.info("Get quote: $id")
+
+        quoteService.getQuote(id)?.let { quote ->
+            quote.toResult(env)
+        } ?: throw QuoteNotFoundQueryException("No quote with id '$id' was found!")
+    }
+
+    fun lastQuoteOfMember(env: DataFetchingEnvironment) {
+
+        logger.info("Get last quote for member")
```
johantjellden, line 41: Perhaps we should extract out the memberId before we log this so we can attach that to the log. That is what we are doing with the `env.getToken()` method below.

johantjellden (commented):
When we get requests by graphql the memberId is implicitly always part of the request.
Left a comment about that and that we should probably log that :)

src/main/kotlin/com/hedvig/underwriter/util/Pii.kt
mightyost: I think the request data is a bit too "unstructured" and too big to annotate each log message sent to DataDog using `MDC`. We should definately send some "standard" markers like `memberId` etc using MDC.

However DataDog does some magic parsing automatically of our free text logs and extracts all `KEY=VALUE` pairs and annotate the log events with it which makes it searchable. That may achieve a bit what you are looking for...

https://docs.datadoghq.com/logs/log_collection/java/?tab=slf4j#using-the-key-value-parser

src/main/kotlin/com/hedvig/underwriter/util/Pii.kt
mightyost: Good point. But as you say, I do not think we need to care about performance optimisations here so let postpone that.

src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
mightyost: Yeah, realised that. It is fixed in commit no 2 :).

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
```diff
@@ -37,6 +39,9 @@ class Mutation @Autowired constructor(
 ) : GraphQLMutationResolver {
 
     fun createQuote(createQuoteInput: CreateQuoteInput, env: DataFetchingEnvironment): CreateQuoteResult {
+
+        logger.info("Create quote: ${createQuoteInput.toNonPiiString()}")
```
vonElfvin, line 43: We use the word "masked" sometimes, maybe you prefer it 🤷 
```suggestion
        logger.info("Create quote: ${createQuoteInput.toMaskedString()}")
```

src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
```diff
@@ -25,16 +27,27 @@ class Query @Autowired constructor(
...
 ) : GraphQLQueryResolver {
 
-    fun quote(id: UUID, env: DataFetchingEnvironment) = quoteService.getQuote(id)?.let { quote ->
-        quote.toResult(env)
-    } ?: throw QuoteNotFoundQueryException("No quote with id '$id' was found!")
+    fun quote(id: UUID, env: DataFetchingEnvironment) {
+
+        logger.info("Get quote for quoteId=$id")
+
+        quoteService.getQuote(id)?.let { quote ->
```
vonElfvin, line 34:
```suggestion
       return quoteService.getQuote(id)?.let { quote ->
```

src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
```diff
@@ -25,16 +27,27 @@ class Query @Autowired constructor(
...
+            quote.toResult(env)
+        } ?: throw QuoteNotFoundQueryException("No quote with id '$id' was found!")
+    }
+
+    fun lastQuoteOfMember(env: DataFetchingEnvironment) {
```
vonElfvin, line 39: missing return

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
```diff
@@ -79,35 +84,51 @@ class Mutation @Autowired constructor(
         }
     }
 
-    fun editQuote(input: EditQuoteInput, env: DataFetchingEnvironment): CreateQuoteResult =
-        responseForEditedQuote(
+    fun editQuote(input: EditQuoteInput, env: DataFetchingEnvironment): CreateQuoteResult {
+
+        logger.info("Edit quote: ${input.toNonPiiString()}")
```
vonElfvin, line 89: If we're gonna use this multiple times in the same way, I'd prefer doing something more generically usable. Maybe have a logger util that takes the class name and function name as a String. Also, variables as a map of String to Any or something 🤷 
`logCall(Mutation::class.simpleName, "editQuote", mapOf("input" to input))` and then log it

src/main/kotlin/com/hedvig/underwriter/graphql/type/BundledQuote.kt
```diff
@@ -1,21 +1,22 @@
...
 import com.hedvig.graphql.commons.type.MonetaryAmountV2
+import com.hedvig.underwriter.util.Pii
 import java.time.LocalDate
 import java.util.UUID
 
 data class BundledQuote(
     val id: UUID,
-    val firstName: String,
-    val lastName: String,
+    @Pii val firstName: String,
```
vonElfvin, line 10: Maybe? 😸 
```suggestion
    @Maskable val firstName: String,
```

src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
```diff
@@ -203,24 +204,35 @@ class SignServiceImpl(
...
+
+        val response = quotes
+            .toOption()
+            .toEither { ErrorResponseDto(ErrorCodes.NO_SUCH_QUOTE, "No such quote $quoteId") }
+            .map { updateNameFromRequest(it, body) }
+            .map { updateEmailFromRequest(it, body) }
+            .map { updateSsnFromRequest(it, body) }
+            .map { updateStartTimeFromRequest(it, body) }
+            .map(::assertAgreementIdIsNull)
+            .map(::assertSsnIsNotNull)
```
vonElfvin, line 221: Weird that we assert after updating 🤷

vonElfvin (commented):
Neat idea, added some suggestions that could maybe make this a bit more easy to reuse and understand?

src/main/kotlin/com/hedvig/underwriter/util/Pii.kt
bystam: Hmm - this is not for performance though. It's rather that the wrong method will be called:
```kotlin
val map = mapOf("key", "value")
map.toNonPiiString() // this will call the  `Map<*, *>` version
val anyMap: Any = map
anyMap.toNonPiiString() // this will call the `Any` version - which will miss the custom formatting for Maps
```

src/main/kotlin/com/hedvig/underwriter/util/Pii.kt
```diff
@@ -0,0 +1,39 @@
...
+    if (obj::class.java.packageName.startsWith("java"))
+        return obj.toString()
+
+    while (clazz != null) {
+        for (prop in clazz.declaredFields.filterNot { Modifier.isStatic(it.modifiers) }) {
+
+            prop.isAccessible = true
+
+            val pii = prop.getAnnotation(Pii::class.java)
+            val value = if (pii != null) "***" else prop.get(obj)?.toNonPiiString()?.trim()
```
bystam, line 32: Do we really want `trim()` here? It feels like it could actually strip away whitespace for "real data".

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
mightyost: Thanks for good comments!

I will change `toMaskedString()` and to `@Masked`.

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
mightyost: Yeah, good point, I will keep this back in my mind if this pattern is to repeat elsewhere.

src/main/kotlin/com/hedvig/underwriter/util/Pii.kt
bystam: I think this is also important due to this part below: `prop.get(obj)?.toNonPiiString()`. That will be called on an `Any`, which could make it lose the information about which implementation it should use:
```kotlin
data class PeopleDatabase(
  val data: Map<String, Person>
)

val database = PeopleDatabase(mapOf("id" to Person("Fredrik Tõnisson-Bystam"))
// the initial call will take the right path, but the subequent call to `prop.get(obj)?.toNonPiiString()`
// on the `data` field will end up at the `.startsWith("java") and thus print the Person without its
// fullName masked
database.toNonPiiString()
```

src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
mightyost: Hmm, cannot add the return, function returns Unit so IntelliJ throws in its veto.

src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
mightyost: Yeah, the `assertAgreementIdIsNull` could be done earlier. This will look a bit different when I refactor this method to support a list of quotes to sign...

src/main/kotlin/com/hedvig/underwriter/util/Pii.kt
mightyost: Haha, I got this mixed up.

Did not see this, good point, will look into it!

mightyost: Fixed together with Elvins re-naming in new commit.

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
mightyost: Fixed in new commit.

src/main/kotlin/com/hedvig/underwriter/util/Pii.kt
bystam: Awesome! 🎉

src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
vonElfvin: I think that is due to it not being specified when using a curly bracket instead of the equals operator. A Query is a getter and we should return something here and below for sure.

Specify the return type (not Unit) and it should work

src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
vonElfvin: This with a if multiple valies makes sense to me

src/main/kotlin/com/hedvig/underwriter/graphql/Query.kt
mightyost: Aha, that was a creepy one! Thanks for seeing that!!

bystam (approved):
This looks great! If we find this to work well, maybe we should lift it to some kind of "hedvig-loggin" library.

PR #201, opened 2021-03-11
Feature/add bbrid to danish quote

AlexSJ1990:
# Jira Issue: [] 

## What?
- Pass bbrId (unique identifier for an apartment/home) that we get from dawa autocomplete service - to price-engine via quote. This id can then be used to call the geomatic api to get further data points for pricing.

- The thinking at the moment is that we will have autocomplete in the web but may not have it in h.ope when we release this - therefore if we change the street or zipcode on the quote we will not default to the previous quote's bbrId but instead set it to null (as the previous bbrId will no longer be relevant and we won't fetch the new one based on the updated address)

## Why?
- enable more accurate pricing.

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [ ] Tested locally


src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
```diff
@@ -315,7 +315,9 @@ data class Quote(
                     firstName = quoteRequest.firstName ?: data.firstName,
                     lastName = quoteRequest.lastName ?: data.lastName,
                     email = quoteRequest.email ?: data.email,
-                    phoneNumber = quoteRequest.phoneNumber ?: data.phoneNumber
+                    phoneNumber = quoteRequest.phoneNumber ?: data.phoneNumber,
+//                    check here
```
bystam, line 319: Is this comment meant to be here?

bystam: (I love how reading my comment alone makes it sound super existential)

src/main/resources/db/changelog/migrations/20210310-1453-add-bbrId-to-danish-home-contents.sql
```diff
@@ -0,0 +1,8 @@
+--liquibase formatted.sql
+
+--changeset alexsj1990:20210310-1453-add-bbrId-to-danish-home-contents.sql
+
+ALTER TABLE quote_revision_danish_home_contents_data
+    ADD COLUMN bbrId text;
```
bystam, line 6: I think it's a common convention to keep SQL type keywords uppercased as well (`TEXT` in this instance).

Alos, do we usually tend to use `text` as the column type for things that are, I'm assuming, non-variable length? An ID sounds like something that would probably be the same length all the time, and then it could probably be safely stored inside something like a `VARCHAR(128)` or such.

bystam (commented):
Looks nice! Added a few small comments/thoughts

src/main/resources/db/changelog/migrations/20210310-1453-add-bbrId-to-danish-home-contents.sql
mightyost, line 6: Regarding:
```
ALTER TABLE quote_revision_danish_home_contents_data
     ADD COLUMN bbrId text;
```

I do not know the default behaviour in Postgres but I guess `bbrId` should be a nullable field and if so I think it is good to state that explicity to avoid misconception. E.g.:
```
ALTER TABLE quote_revision_danish_home_contents_data
     ADD COLUMN bbrId text NULL;
```

And maybe `bbrId` should be a `varchar` instead, it is much more conventional than using `text`...

And looking at the naming conventions of other columns it looks like `bbrId` rather should be named `bbrid`, i.e. all lowercase to be aligned.

src/main/resources/db/changelog/migrations/20210310-1453-add-bbrId-to-danish-home-contents.sql
```diff
@@ -0,0 +1,8 @@
...
+
+--rollback ALTER TABLE quote_revision_danish_home_contents_data DROP COLUMN bbrId text;
```
mightyost, line 8: I think:
```
--rollback ALTER TABLE quote_revision_danish_home_contents_data DROP COLUMN bbrId text;
```
...should be just:
```
--rollback ALTER TABLE quote_revision_danish_home_contents_data DROP COLUMN bbrId;
```

src/main/resources/schema.graphqls
```diff
@@ -62,6 +62,7 @@ input CreateNorwegianTravelInput {
 input CreateDanishHomeContentsInput {
     street: String!
     zipCode: String!
+    bbrId: String
```
mightyost, line 65: I do know nothing about graphql. Why is `bbrId` not nullable here but everywhere else?

src/main/kotlin/com/hedvig/underwriter/graphql/type/QuoteDetails.kt
```diff
@@ -76,6 +76,7 @@ sealed class QuoteDetails {
     data class DanishHomeContentsDetails(
         @Masked val street: String,
         val zipCode: String,
+        val bbrId: String?,
```
mightyost, line 79: If `bbrId` is a PII sensitive field (i.e. can be used to identify a person) then we should add the `@Masked` to it everywhere to omit it from being written out in our logs?

src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
mightyost, line 319: I do not know if this is intentional or just forgotten:
```
// check here
```

src/main/resources/schema.graphqls
AlexSJ1990: this is how you show it is nullable in graphql - the syntax would be ```bbrId: String!``` (if it was non-nullable)

src/main/resources/db/changelog/migrations/20210310-1453-add-bbrId-to-danish-home-contents.sql
AlexSJ1990: 👍

src/main/resources/db/changelog/migrations/20210310-1453-add-bbrId-to-danish-home-contents.sql
AlexSJ1990: 👍

AlexSJ1990: good point, will make it explicit that it can be null and amend to varchar

src/main/kotlin/com/hedvig/underwriter/graphql/type/QuoteDetails.kt
AlexSJ1990: good point, I am slightly unsure if it is or not -> it can be used to identify a building but don't think any info relating to a person

src/main/resources/schema.graphqls
mightyost: I have much to learn :)!

src/main/kotlin/com/hedvig/underwriter/graphql/type/QuoteDetails.kt
mightyost: If it can identify eg a single apartment or villa then we should mask it. Better safe than sorry. I am masking `street address` for the same reason...

AlexSJ1990: makes sense - will mask :)

src/test/kotlin/com/hedvig/underwriter/service/CreateQuoteTest.kt
```diff
@@ -50,6 +55,31 @@ class CreateQuoteTest {
...
+
+        every { strategyService.getAllGuidelines(any()) } returns setOf()
+        every { priceEngineService.queryDanishHomeContentPrice(capture(danishHomeContentData)) } returns PriceQueryResponse(
+            UUID.randomUUID(),
+            Money.of(1, "NOK")
+        )
+        every { quoteRepository.insert(any(), any()) } returns Unit
+
+        val result = cut.createQuote(request, UUID.randomUUID(), QuoteInitiatedFrom.ANDROID, null, true)
+        assert(result is Either.Right)
```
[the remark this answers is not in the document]
AlexSJ1990: the only way i could get it to compile and run with with assertThat was to do this but slightly unsure about it ```assertThat(result).isInstanceOf(Either.Right.right().javaClass)```

mightyost (approved):
Looks good :)!

PR #202, opened 2021-03-12
Support for logging request calls in a standardised and nonintrusive way

mightyost:
## What?
Added support for logging request calls in a simple, standardised, nonintrusive and gdpr safe way using spring-aop.

By annotating a method with `@LogCall` will result in calls to it to be logged out with input parameters as well as returned value. Logged objects data will be masked out for fields marked with the `@Masked` annotation.

E.g. a call to this:
```
    @PostMapping
    @LogCall
    fun createQuote(
        @Valid @RequestBody requestDto: QuoteRequestDto,
        httpServletRequest: HttpServletRequest
    ): ResponseEntity<out Any>
```
...will result in following logs:
```
Calling QuoteController.createQuote(..), parameters: [QuoteRequestDto(firstName=null, lastName=null, email=null, currentInsurer=null, birthDate=1988-01-01, ssn=null, quotingPartner=HEDVIG, productType=HOME_CONTENT, incompleteQuoteData=NorwegianHomeContents(street=***, zipCode=1234, city=ApCity, coInsured=1, livingSpace=122, isYouth=false, subType=OWN)), ch.qos.logback.access.servlet.TeeHttpServletRequest@34892a6b]
...

Finished executing: QuoteController.createQuote(..), returned: '<200,CompleteQuoteResponseDto(id=a273062c-0465-47bf-9aac-2108324582ae, price=200, currency=NOK, validTo=2021-04-11T07:19:15.122856Z),[]>', duration: 40 ms
```

All the magic is done in the new small `LogCallAspect` class. The rest of the changes in this PR is just to incorporate the `@LogCall` and `@Masked` annotation into underwriter service.

There is room for improvements, e.g. I could not manage to make `@LogCall` work on class level which would be very convenient. This due to some strange aop error complaining at a LiquidBaseConfig thingy.

After battle testing this in the underwriter service and get it right I think this logic should be extracted into a sharable lib for other services to start using.

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/util/logging/LogCallAspect.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/util/logging/LogCallAspect.kt
@@ -1,0 +1,33 @@
+package com.hedvig.underwriter.util.logging
+
+import com.hedvig.underwriter.util.logger
+import org.aspectj.lang.ProceedingJoinPoint
+import org.aspectj.lang.annotation.Around
+import org.aspectj.lang.annotation.Aspect
+import org.springframework.stereotype.Component
+
+@Aspect
+@Component
+class LogCallAspect {
+
+    @Suppress("TooGenericExceptionCaught")
+    @Around("@annotation(com.hedvig.underwriter.util.logging.LogCall)")
+    fun logExecutionTime(joinPoint: ProceedingJoinPoint): Any {
+        val start = System.currentTimeMillis()
+        val signature = joinPoint.signature.toShortString()
+        val result = try {
+            logger.info("Calling $signature, parameters: [${getParams(joinPoint.args)}]")
+            joinPoint.proceed()
+        } catch (throwable: Throwable) {
+            logger.error("Exception during executing $signature,", throwable)
+            throw throwable
+        }
+        val duration = System.currentTimeMillis() - start
+        logger.info("Finished executing: $signature, returned: '$result', duration: $duration ms")
+        return result
+    }
+
+    private fun getParams(args: Array<Any>) =
+        args.map { it.toMaskedString() }
+            .joinToString(", ")
+}
--- a/src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/graphql/Mutation.kt
@@ -23,8 +23,7 @@ import com.hedvig.underwriter.model.QuoteInitiatedFrom
 import com.hedvig.underwriter.service.QuoteService
 import com.hedvig.underwriter.service.SignService
 import com.hedvig.underwriter.service.model.StartSignResponse
-import com.hedvig.underwriter.util.logger
-import com.hedvig.underwriter.util.toMaskedString
+import com.hedvig.underwriter.util.logging.LogCall
 import com.hedvig.underwriter.web.dtos.ErrorCodes
 import com.hedvig.underwriter.web.dtos.ErrorResponseDto
 import graphql.schema.DataFetchingEnvironment
@@ -39,9 +38,8 @@ class Mutation @Autowired constructor(
     private val quoteMapper: QuoteMapper
 ) : GraphQLMutationResolver {
 
+    @LogCall
     fun createQuote(createQuoteInput: CreateQuoteInput, env: DataFetchingEnvironment): CreateQuoteResult {
-
-        logger.info("Create quote: ${createQuoteInput.toMaskedString()}")
 
         val input = when {
             createQuoteInput.apartment != null || createQuoteInput.house != null ||
@@ -85,9 +83,8 @@ class Mutation @Autowired constructor(
         }
     }
 
+    @LogCall
     fun editQuote(input: EditQuoteInput, env: DataFetchingEnvironment): CreateQuoteResult {
-
-        logger.info("Edit quote: ${input.toMaskedString()}")
 
         return responseForEditedQuote(
             quoteService.updateQuote(
@@ -98,9 +95,8 @@ class Mutation @Autowired constructor(
         )
     }
 
+    @LogCall
     fun removeCurrentInsurer(input: RemoveCurrentInsurerInput, env: DataFetchingEnvironment): CreateQuoteResult {
-
-        logger.info("Remove current insurance: ${input.toMaskedString()}")
 
         return responseForEditedQuote(
             quoteService.removeCurrentInsurerFromQuote(input.id),
@@ -108,9 +104,8 @@ class Mutation @Autowired constructor(
         )
     }
 
+    @LogCall
     fun removeStartDate(input: RemoveStartDateInput, env: DataFetchingEnvironment): CreateQuoteResult {
-
-        logger.info("Remove current insurance: ${input.toMaskedString()}")
 
         return responseForEditedQuote(
             quoteService.removeStartDateFromQuote(input.id),
@@ -118,9 +113,8 @@ class Mutation @Autowired constructor(
         )
     }
 
+    @LogCall
     fun signQuotes(input: SignQuotesInput, env: DataFetchingEnvironment): StartSignResponse {
-
-        logger.info("Sign quotes: ${input.toMaskedString()}")
 
         return signService.startSigningQuotes(
             input.quoteIds,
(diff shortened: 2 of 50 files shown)


src/main/kotlin/com/hedvig/underwriter/util/logging/LogCallAspect.kt
```diff
@@ -0,0 +1,33 @@
...
+        val result = try {
+            logger.info("Calling $signature, parameters: [${getParams(joinPoint.args)}]")
+            joinPoint.proceed()
+        } catch (throwable: Throwable) {
+            logger.error("Exception during executing $signature,", throwable)
+            throw throwable
+        }
+        val duration = System.currentTimeMillis() - start
+        logger.info("Finished executing: $signature, returned: '$result', duration: $duration ms")
+        return result
```
bystam, line 27: Instead of using `com.hedvig.underwriter.util.logger`, would it make sense to create a `Logger` from the `joinPoint.signature.declaringType`? That would mean the logger name of the statement matches the type itself, instead of all of them being printed by `LogCallAspect`. I like using services like Datadog to sometimes filter by a single type to see all the logs happening for it.

bystam (commented):
This looks *really* nice! And way more stable than having to remember to do `toMaskedString()` oneself all the time. This should probably be part of a library that can be dropped into more services. :) 

Just to propose a possible alternative for discussions sake. What does our risk profile look like when it comes to sensitive data slipping into the logs? Have we considered doing the exact same thing as `@Masked` but **flipped**? As in, assuming all values should be masked by default, and instead actively "revealing" values that aren't? Something like a `@SafelyLogged`? That would probably flatten the risk of logging sensitive data way closer to zero, at the cost of maybe losing some information by default.

But my calibration for this risk may be way off, since it's based on my previous job where lots of the data was arguably **much** more sensitive (medical data that is).

src/main/kotlin/com/hedvig/underwriter/util/logging/LogCallAspect.kt
```diff
@@ -0,0 +1,33 @@
...
+@Component
+class LogCallAspect {
+
+    @Suppress("TooGenericExceptionCaught")
+    @Around("@annotation(com.hedvig.underwriter.util.logging.LogCall)")
+    fun logExecutionTime(joinPoint: ProceedingJoinPoint): Any {
+        val start = System.currentTimeMillis()
+        val signature = joinPoint.signature.toShortString()
+        val result = try {
+            logger.info("Calling $signature, parameters: [${getParams(joinPoint.args)}]")
```
vonElfvin, line 19: Is there any concept like a "thread id" or like "request id" that we can access to know of the "chain of calls" made, for debugging purposes?

AlexSJ1990 (commented):
I think this looks awesome and is a great idea! Feels like we are moving in a great direction!

The only thing that I wonder about is, when we search for logs in Datadog it is really useful to be able to search via memberId (in a lot of cases). It would be amazing if we could always capture the member id in the logs without necessarily having to pass it in. Think this is completely out of scope for this pr but just wanted to raise it as we are thinking about the future of logging in all of our services.

src/main/kotlin/com/hedvig/underwriter/util/logging/LogCallAspect.kt
mightyost: Yes, I had it like that in the first place but changed to this. Thought it was maybe confusing for people not to find the actual logging instruction in class since that what you should expect. But I agree with you, will change back. Or maybe to:
```
joinPoint.signature.declaringType + "-aop"
```
...will that work for you in Datadog?

src/main/kotlin/com/hedvig/underwriter/util/logging/LogCallAspect.kt
mightyost: That is handled by the logging framework. This is how it really looks right now when I run a test case:
```
2021-03-12 08:19:15.286  INFO [underwriter,3fb3cbc1da0691ff,3fb3cbc1da0691ff,false] 40172 --- [o-auto-1-exec-4] c.h.u.util.logging.LogCallAspect         : Calling QuoteController.signQuotesFromRapio(..), parameters: [SignQuotesRequestDto(quoteIds=[012e7567-8c0b-4249-9a0c-fab30e4ef46a, a273062c-0465-47bf-9aac-2108324582ae], name=Name(firstName=***, lastName=***), ssn=***, startDate=2021-03-12, email=***, price=150.0, currency=NOK)]
2021-03-12 08:19:15.318  INFO [underwriter,3fb3cbc1da0691ff,3fb3cbc1da0691ff,false] 40172 --- [o-auto-1-exec-4] c.h.underwriter.service.SignServiceImpl  : New member created: 7368445544159342915
2021-03-12 08:19:15.508  INFO [underwriter,3fb3cbc1da0691ff,3fb3cbc1da0691ff,false] 40172 --- [o-auto-1-exec-4] c.h.u.util.logging.LogCallAspect         : Finished executing: QuoteController.signQuotesFromRapio(..), returned: '<200,SignedQuotesResponseDto(memberId=7368445544159342915, market=NORWAY, contracts=[Contract(id=1a0211ed-0262-4ede-9eba-5be6bca04355, signedAt=2021-03-12T07:19:15.388582Z), Contract(id=1a0211ed-0262-4ede-9eba-5be6bca04355, signedAt=2021-03-12T07:19:15.460784Z)]),[]>', duration: 221 ms
```
Here the `o-auto-1-exec-4` is the thread name. You also see other things like `traceId` and can add custom ones like e.g. `memberId` through `MDC` if we want...

mightyost:
> The only thing that I wonder about is, when we search for logs in Datadog it is really useful to be able to search via memberId (in a lot of cases). It would be amazing if we could always capture the member id in the logs without necessarily having to pass it in. Think this is completely out of scope for this pr but just wanted to raise it as we are thinking about the future of logging in all of our services.

Totally agree. I think we can do this in DataDog already, they have quite powerful tooling to setup parsing rules and indexing on parsed outcome in the logging pipelines. But I also think a free text search for e.g. a memberId should work quite well...

src/main/kotlin/com/hedvig/underwriter/util/logging/LogCallAspect.kt
bystam: I think that should work, but I think I would prefer a prefix rather than a suffix, because there are already prefixes to types (the package name).

If I want to search for `QuoteController`, then I would right now (I think? I am used to Kibana and Lucene here) do something like this:
```
logger_name:*QuoteController
```
if we added `-aop` at the end it would become 
```
logger_name:*QuoteController*
```
which I think miiiight cause slower searching sometimes? I wonder if searching in datadog works anything like searching with `LIKE` in SQL where it's fast as long as it's either a prefix or a suffix but both? I am not super sure here. Maybe I'm just completely wrong.

-----

Regardless, regarding this:
> Thought it was maybe confusing for people not to find the actual logging instruction in class since that what you should expect.
To me that should be a passing issue during initial rollout only. If people get used to this library-style functionality then it shouldn't be that confusing after a while, no?


-----

Another idea! We could do a hybrid. Use this class as the regular logger name, but add a structured value (using MDC or Markers) for `declaring_type` or something. That way it would hit when searching too, just on another field.

src/main/kotlin/com/hedvig/underwriter/util/logging/LogCallAspect.kt
bystam: That's what you usually use traceId headers for. I think amazon attaches them automatically to requests. Do you know if this works already @mightyost ?

bystam:
> > The only thing that I wonder about is, when we search for logs in Datadog it is really useful to be able to search via memberId (in a lot of cases). It would be amazing if we could always capture the member id in the logs without necessarily having to pass it in. Think this is completely out of scope for this pr but just wanted to raise it as we are thinking about the future of logging in all of our services.
> 
> Totally agree. I think we can do this in DataDog already, they have quite powerful tooling to setup parsing rules and indexing on parsed outcome in the logging pipelines. But I also think a free text search for e.g. a memberId should work quite well...

In my opinion, what I would want is an interceptor on all requests that attaches the member_id as a structured value to the MDC using an interceptor. We already know that the `hedvig.token` header usually contains that value, so if we simply attach it to literally all requests, that should give us great coverage immediately, and make it become a value in Datadog that can be both searched for, but also shown as its own column in the log stream.

Something like this:
http://www.devgrok.com/2017/04/adding-mdc-headers-to-every-spring-mvc.html

bystam:
If we did something like:
```kotlin
MDC.put(
  "hedvig",
  mapOf("memberId" to memberId
)
```

then we would get some pretty awesome data in Datadog where we can see our own specific values grouped by a nice prefix :)

bystam (approved):
Awesome!

PR #203, opened 2021-03-15
Masking the method result logged in `LogAspect`.

mightyost:
## What?
The method results are now logged with pin masking which was forgotten.

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


bystam (approved):
Good catch!

PR #204, opened 2021-03-15
Handle NPE gracefully in `QuoteController.getMarketInfoFromLatestQuote`

mightyost:
We got some NPEs in dev due to quote not found for a member, this commit will handle it more gracefully.

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -162,7 +163,7 @@ class QuoteServiceImpl(
     }
 
     override fun getMarketFromLatestQuote(memberId: String): Market {
-        return getLatestQuoteForMemberId(memberId)!!.market
+        return getLatestQuoteForMemberId(memberId)?.market ?: throw IllegalArgumentException("Member $memberId has no quotes")
```
palmenhq, line 166: This is very nit but i'd argue this is due to an illegal state, not an illegal argument, or?

PR #205, opened 2021-03-15
Fix skip after guideline hit

FredrikAreschoug:
## What?
- Fix for skip logic after guideline hit

## Why?
- I saw something strange when going through this code with @mightyost.

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [ ] Tested locally

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/service/UnderwriterImpl.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/service/UnderwriterImpl.kt
@@ -143,13 +143,15 @@ class UnderwriterImpl(
     ): MutableList<BreachedGuidelineCode> {
         val guidelineErrors = mutableListOf<BreachedGuidelineCode>()
 
-        listOfRules.forEach {
-            val error = it.invokeValidate(data)
-            if (error != null) {
-                guidelineErrors.add(error)
-            }
-            if (it.skipAfter) {
-                return@forEach
+        run validateGuidelines@{
+            listOfRules.forEach {
+                val error = it.invokeValidate(data)
+                if (error != null) {
+                    guidelineErrors.add(error)
+                    if (it.skipAfter) {
+                        return@validateGuidelines
+                    }
+                }
             }
         }
         return guidelineErrors
--- a/src/main/kotlin/com/hedvig/underwriter/service/UnderwriterImpl.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/service/UnderwriterImpl.kt
@@ -143,14 +143,12 @@ class UnderwriterImpl(
     ): MutableList<BreachedGuidelineCode> {
         val guidelineErrors = mutableListOf<BreachedGuidelineCode>()
 
-        run validateGuidelines@{
-            listOfRules.forEach {
-                val error = it.invokeValidate(data)
-                if (error != null) {
-                    guidelineErrors.add(error)
-                    if (it.skipAfter) {
-                        return@validateGuidelines
-                    }
+        for (rule in listOfRules) {
+            val error = rule.invokeValidate(data)
+            if (error != null) {
+                guidelineErrors.add(error)
+                if (rule.skipAfter) {
+                    break
                 }
             }
         }
--- a/src/test/kotlin/com/hedvig/underwriter/service/UnderwriterImplValidateAndCompleteQuoteTest.kt
+++ b/src/test/kotlin/com/hedvig/underwriter/service/UnderwriterImplValidateAndCompleteQuoteTest.kt
@@ -90,9 +90,7 @@ object FailsShouldNotSkipAfter : BaseGuideline<QuoteData> {
     override val skipAfter: Boolean
         get() = false
 
-    override val validate = { data: QuoteData ->
-        true
-    }
+    override val validate : (QuoteData)-> Boolean = { true }
 }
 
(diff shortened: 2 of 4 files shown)


src/main/kotlin/com/hedvig/underwriter/service/UnderwriterImpl.kt
```diff
@@ -143,13 +143,15 @@ class UnderwriterImpl(
...
-                return@forEach
+        run validateGuidelines@{
+            listOfRules.forEach {
+                val error = it.invokeValidate(data)
+                if (error != null) {
+                    guidelineErrors.add(error)
+                    if (it.skipAfter) {
+                        return@validateGuidelines
+                    }
+                }
```
bystam, line 154: Wouldn't it be simpler to read if this were just a:
```suggestion
        for (rule in listOfRules) {
            val error = it.invokeValidate(data)
            if (error != null) {
                guidelineErrors.add(error)
                if (it.skipAfter) {
                    break
                }
            }
```
I feel the extra `run` and labeling feels like a double-dance to avoid a simple for loop?

src/test/kotlin/com/hedvig/underwriter/service/UnderwriterImplValidateAndCompleteQuoteTest.kt
```diff
@@ -0,0 +1,121 @@
...
+
+class UnderwriterImplValidateAndCompleteQuoteTest {
+
+    val quoteStrategyService: QuoteStrategyService = mockk()
+
+    val cut = UnderwriterImpl(
+        mockk(relaxed = true),
+        quoteStrategyService,
+        mockk(relaxed = true)
+    )
```
bystam, line 21: Making this not re-created in a `@BeforeEach` could make mocks from one test bleed into the next. Maybe that's safe here though?

src/test/kotlin/com/hedvig/underwriter/service/UnderwriterImplValidateAndCompleteQuoteTest.kt
```diff
@@ -0,0 +1,121 @@
...
+}
+
+object SuccessShouldNotSkipAfter : BaseGuideline<QuoteData> {
+    override val breachedGuideline = this::class.simpleName!!
+
+    override val skipAfter: Boolean
+        get() = false
+
+    override val validate : (QuoteData)-> Boolean = { false }
+}
```
bystam, line 121: maybe these should be either nested or `private` to not bleed into global auto complete scope? :)

bystam (commented):
Looks good, just some small thoughts

src/test/kotlin/com/hedvig/underwriter/service/UnderwriterImplValidateAndCompleteQuoteTest.kt
```diff
@@ -0,0 +1,121 @@
...
+import com.hedvig.underwriter.service.guidelines.BaseGuideline
+import com.hedvig.underwriter.service.quoteStrategies.QuoteStrategyService
+import com.hedvig.underwriter.testhelp.databuilder.QB
+import io.mockk.every
+import io.mockk.mockk
+import org.junit.jupiter.api.Test
+
+class UnderwriterImplValidateAndCompleteQuoteTest {
+
+    val quoteStrategyService: QuoteStrategyService = mockk()
```
bystam, line 15: I like using `mockk()` over the annotations 👍. Pretty sure I've read that that's subsantially faster than scanning with reflection.

src/main/kotlin/com/hedvig/underwriter/service/UnderwriterImpl.kt
FredrikAreschoug: Agreed the @ concept is maybe not the best :slightly_smiling_face:

src/test/kotlin/com/hedvig/underwriter/service/UnderwriterImplValidateAndCompleteQuoteTest.kt
FredrikAreschoug: For earlier versions of JUnit that is true but JUnit5 creates a new instance of the test class on each run so I would say that we are safe here. [It is mentioned here](https://junit.org/junit5/docs/current/user-guide/#writing-tests-test-instance-lifecycle)

src/test/kotlin/com/hedvig/underwriter/service/UnderwriterImplValidateAndCompleteQuoteTest.kt
FredrikAreschoug: Yes I'm on the same page as you :slightly_smiling_face:

src/test/kotlin/com/hedvig/underwriter/service/UnderwriterImplValidateAndCompleteQuoteTest.kt
bystam: Oh wow - today I learned! I'm sorry! :D I was confused with the naming of `@BeforeEach` here too, but they even mention that in the guide you linked.

mightyost (approved):
Gr8!

Michael-Duivestein (approved):

bystam (approved):

PR #208, opened 2021-03-17
Feature/enable simple sign in denmark

FredrikAreschoug:
## What?
- Enable simple sign in denmark
- Clean up of redirect

## Why?
- We want to use simple sign in denmark

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/SignStrategyService.kt
```diff
@@ -28,7 +27,7 @@ class SignStrategyService(
             return StartSignErrors.noQuotes
         }
 
-        if (strategy.size > 1) {
+        if (strategy.size > 1 || quotes.markets.size > 1) {
```
vonElfvin, line 30: Can't we just add a new StartSignError?

src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/RedirectSignStrategy.kt
```diff
@@ -1,88 +0,0 @@
...
-import com.hedvig.underwriter.serviceIntegration.memberService.dtos.NationalIdentification
-import com.hedvig.underwriter.serviceIntegration.memberService.dtos.RedirectCountry
-import com.hedvig.underwriter.serviceIntegration.memberService.dtos.UnderwriterStartSignSessionResponse
-import com.hedvig.underwriter.serviceIntegration.memberService.dtos.Nationality
-import org.springframework.stereotype.Service
-import java.util.UUID
-import kotlin.IllegalStateException
-
-@Service
-class RedirectSignStrategy(
```
vonElfvin, line 19: Would we need this if we went to a new market anytime soon?

vonElfvin (approved):
Looks good, added two questions that are maybe worth considering

src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/RedirectSignStrategy.kt
FredrikAreschoug: I don't know but you can always find it in the git history

src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/SignStrategyService.kt
FredrikAreschoug: Yes we can but if I don't see why the thing is that the quotes can not be bundled if they are of different markets

PR #209, opened 2021-03-18
MX-137: Wrap the product-pricing selfChange functionality as a SignStrategy

bystam:
# Jira Issue: [MX-137] , [MX-148], [MX-21]

## What?
- Tag quotes with a new `QuoteInitiatedFrom` type (SELF_CHANGE)
- Use the new `initatedFrom` value to triage differently in `signMethodForQuotes`, with a new `SignMethod.APPROVAL_ONLY`
- Writes some actual tests that runs the quote APIs from a HTTP point of view, without mocks (except some external service fakes)

## Why?
- The offer page needs to know that it should "sign" these quotes differently from any of the quotes that are produced in the onboarding flow
- Quotes and C&A need to be connected upon changes
- slack thread: https://hedviginsurance.slack.com/archives/C01NB6GBRFS/p1622013719001900

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [x] Tested locally

[MX-148]: https://hedvig.atlassian.net/browse/MX-148
[MX-21]: https://hedvig.atlassian.net/browse/MX-21

[MX-137]: https://hedvig.atlassian.net/browse/MX-137


Michael-Duivestein (approved):

src/main/kotlin/com/hedvig/underwriter/service/SelfChangeServiceImpl.kt
```diff
@@ -0,0 +1,27 @@
...
+    private val productPricingService: ProductPricingService
+) : SelfChangeService {
+    override fun changeToQuotes(quoteIds: List<UUID>, memberId: String) {
+        val quotes = quoteRepository.findQuotes(quoteIds)
+        val result = productPricingService.selfChangeContracts(
+            memberId = memberId,
+            quotes = quotes
+        )
+        val updates = result.createdContracts + result.updatedContracts
+        updates.forEach { contract ->
```
vonElfvin, line 20: Looking at this line it feels a bit weird with the name "contract", like "contract.quoteId" sounds weird to me

src/main/kotlin/com/hedvig/underwriter/service/SelfChangeServiceImpl.kt
```diff
@@ -0,0 +1,27 @@
...
+            val quote = quotes.first { contract.quoteId == it.id }
+            quoteRepository.update(
+                quote.copy(contractId = contract.contractId, agreementId = contract.agreementId)
```
vonElfvin, line 23: I'm not 100% on this, so please double check:
I think that we only set the contractId **if** if create a contract, otherwise only agreementId

vonElfvin (commented):
Very neat

src/main/kotlin/com/hedvig/underwriter/service/SelfChangeServiceImpl.kt
bystam: Yeah, maybe. It's renamed to `ContractChange` before, so I could rename it `change` here

src/main/kotlin/com/hedvig/underwriter/service/SelfChangeServiceImpl.kt
bystam: It... feels like it should be the same? As in, if therels already a `quote<=>contract` connection then it should have their contractId matching. But maybe it's good to split it up anyway, just to be safe.

src/main/kotlin/com/hedvig/underwriter/service/QuoteService.kt
```diff
@@ -32,6 +32,8 @@ interface QuoteService {
         underwritingGuidelinesBypassedBy: String?
     ): Either<ErrorResponseDto, CompleteQuoteResponseDto>
 
+    fun bindQuoteToContract(quoteId: UUID, contractId: UUID, agreementId: UUID): Either<ErrorResponseDto, Quote>
```
bystam, line 35: I was contemplating whether this should simply be a:
```
fun updateQuote(quote: Quote)
```
but maybe that would enable too many things?

src/test/kotlin/com/hedvig/underwriter/web/ExternalQuoteControllerTest.kt
```diff
@@ -0,0 +1,120 @@
...
+        val validTo: Instant? = null,
+        val breachedUnderwritingGuidelines: List<Any>? = null
+    )
+
+    private data class Quote(
+        val id: UUID,
+        val contractId: UUID? = null,
+        val agreementId: UUID? = null
+    )
+}
```
bystam, line 120: This file to me is what it's all about! It's a `@SpringBootTest` that does not have a single `@MockBean`. It has a "fake configuration", but that one is global to the entire testing context, which means that it will load once and be cached between every test. Booting the application in the test is slow the first time, but beyond that it will be blazing fast.

Also, this test actually tests the public API of the service, and no implementation details.

src/test/kotlin/com/hedvig/underwriter/TestFakesConfiguration.kt
```diff
@@ -0,0 +1,96 @@
...
+import org.springframework.http.ResponseEntity
+
+@Configuration
+class TestFakesConfiguration {
+    @get:Bean
+    val priceEngineClient: FakePriceEngineClient = FakePriceEngineClient()
+
+    @get:Bean
+    val memberServiceClient: FakeMemberServiceClient = FakeMemberServiceClient()
+}
```
bystam, line 30: An improvement to this would be if we can turn it into a `@Rule` thingy and have it reset automatically between tests.

src/main/kotlin/com/hedvig/underwriter/web/ExternalQuoteController.kt
```diff
@@ -42,4 +48,27 @@ class ExternalQuoteController(
...
+        @PathVariable quoteId: UUID,
+        @RequestBody body: UpdateQuoteContractDto
+    ): ResponseEntity<out Any> {
+        return quoteService.bindQuoteToContract(
+            quoteId, body.contractId, body.agreementId
+        ).bimap(
+            { ResponseEntity.status(422).body(it) },
+            { ResponseEntity.status(200).body(it) }
+        ).getOrHandle { it }
+    }
```
mightyost, line 73: I have seen that we return the `Quote` object in other endpoints but I think it is a really bad pattern to return "internal" model objects rather than contracted DTOs. It is very easy to break backward compatibility by misstake and also to "leak" data that should not be exposed externally.

I think the GraphQL endpoints has DTOs for fetching quotes...

src/test/kotlin/com/hedvig/underwriter/web/ExternalQuoteControllerTest.kt
mightyost: Agree and I am all up for it. But instead of fake clients we could also use WireMock or own server stubs to be able to intercept outgoing calls and being able to assert the output.

src/main/kotlin/com/hedvig/underwriter/web/ExternalQuoteController.kt
bystam: Yeah I agree with that - I mostly imitated the existing pattern. I could change it for these while I'm at it.

mightyost (approved):
:)

src/main/kotlin/com/hedvig/underwriter/web/ExternalQuoteController.kt
mightyost: If you have the time for fixing it I think it would be a good "investment".

mightyost: `Quote` is a kind of complex and big object so I can understand why this shortcut was made in the first place ;)...

src/test/kotlin/com/hedvig/underwriter/web/ExternalQuoteControllerTest.kt
bystam: Yeah that could be cool, I have never used it. What are the major strengths of doing something like that over simply just writing regular POJOs? Is it just ergonomics around simulating things such as 400-599 etc?

mightyost: Yes that's one use case. 

But most importantly is that you want to catch the data in the outgoing calls and assert on it, i.e. creating a quote in Underwriter should result in a call to price-engine with all the right data etc.

Also to be able to simulate different responses from called micro services. E.g. test that Underwriter behaves as it should if the call to `member-service.checkPersonDebt(..)` when creating quotes etc...

mightyost: I do not have that much know-how around WireMock (but @fredriklagerblad has), in my previous places we used home made stubs that spun up a small server catching the outgoing calls making it very easy too assert outgoing data and feed back custom responses.

bystam: > But most importantly is that you want to catch the data in the outgoing calls and assert on it, i.e. creating a quote in Underwriter should result in a call to price-engine with all the right data etc.

> Also to be able to simulate different responses from called micro services. E.g. test that Underwriter behaves as it should depending on the different responses from calls to member-service.checkPersonDebt(..) when creating quotes etc...

Right - what I meant is that it's quite easy to just do that with the fakes I built for this PR too. It's just a little manual work putting a setter and all that in there. But Wiremock sounds like a more powerful version of all that. And then one doesn't need to do any faking of beans at all, which I guess is way nicer.

Regardless - I don't think I will apply Wiremock to this PR specifically. Maybe if I have time I'll look into it as the next step.

src/main/kotlin/com/hedvig/underwriter/service/SelfChangeService.kt
```diff
@@ -0,0 +1,31 @@
...
+        val changes = result.createdContracts + result.updatedContracts
+        changes.forEach { change ->
+            val quote = quotes.first { change.quoteId == it.id }
+            quoteRepository.update(
+                quote.copy(contractId = change.contractId, agreementId = change.agreementId, state = QuoteState.SIGNED)
+            )
+        }
+        return result
+    }
+}
```
bystam, line 31: Adding one-off services that wrap other services to me is not a super nice pattern, but I think I had to do it here.

src/test/kotlin/com/hedvig/underwriter/testhelp/TestHttpClient.kt
```diff
@@ -0,0 +1,72 @@
...
+        inline fun <reified T> body(): T {
+            return body(T::class.java)
+        }
+
+        @PublishedApi
+        internal fun <T> body(type: Class<T>): T {
+            return mapper.convertValue(entity.body, type)
+        }
+    }
+}
```
bystam, line 72: The `TestRestTemplate` class has a not very fun API - creating a little wrapper like this makes it way nicer imo to write reusable functions in tests like this: https://github.com/HedvigInsurance/underwriter/pull/209/files#diff-f485f3e5aea5f1f0932c88f6a0a4ae1e4e05a5d9980983b88aa7b12890785373R41

src/test/kotlin/com/hedvig/underwriter/web/ExternalQuoteControllerTest.kt
```diff
@@ -0,0 +1,160 @@
...
+        assertThat(result.updatedContractIds).isEqualTo(listOf(cid1))
+        assertThat(result.createdContractIds).isEqualTo(listOf(cid3))
+        assertThat(result.terminatedContractIds).isEqualTo(listOf(cid2))
+
+        assertThat(updatedQ1.contractId).isEqualTo(cid1)
+        assertThat(updatedQ2.contractId).isEqualTo(cid3)
+    }
+
+    private fun createQuote(
+        bodyChanges: CreateQuoteInput.() -> Unit = {}
```
mightyost, line 102: There exists convenient methods for creating all kind of quotes in the `com.hedvig.underwriter.testhelp.QuoteClient`. (However that class should probably be migrated to use your `TestHttpClient` wrapper since it will make it a bit more easy to work with...)

src/test/kotlin/com/hedvig/underwriter/testhelp/TestHttpClient.kt
```diff
@@ -0,0 +1,72 @@
...
+import org.springframework.boot.test.web.client.TestRestTemplate
+import org.springframework.http.HttpEntity
+import org.springframework.http.HttpHeaders
+import org.springframework.http.HttpMethod
+import org.springframework.http.HttpStatus
+import org.springframework.http.ResponseEntity
+import org.springframework.stereotype.Component
+
+@Component
+class TestHttpClient(
```
mightyost, line 16: Nice class!

src/main/kotlin/com/hedvig/underwriter/web/ExternalQuoteController.kt
```diff
@@ -42,4 +50,44 @@ class ExternalQuoteController(
...
+    @GetMapping("/{id}")
+    @LogCall
+    fun getQuote(
+        @PathVariable id: UUID
+    ): ResponseEntity<QuoteDto> {
+        val quote = quoteService.getQuote(id) ?: return ResponseEntity.notFound().build()
+        return ResponseEntity.ok(QuoteDto.from(quote))
+    }
+
+    @PostMapping("/selfChange")
```
mightyost, line 63: For me it is hard to understand what the purpose of this endpoint is looking at the name `selfChange`, I do not know if it could be changed into something more crisp and self explanational... Looks more like `connectContracts` or something like it?

mightyost (approved):
Looks good!

I like that you also are code scouting a bit with the fake clients.

src/main/kotlin/com/hedvig/underwriter/web/ExternalQuoteController.kt
bystam: Right... that makes sense. The name is kind of terrible. The reason I didn't invent anything new was mostly to make sure the same name is used everywhere (giraffe, product-pricing etc). But I guess that's a quite monolithic mindset.

I will revisit the name quickly. Maybe `/contracts/flushToContracts`? I think the word `connect` implies that the contracts already exist, which they don't yet.

src/test/kotlin/com/hedvig/underwriter/web/ExternalQuoteControllerTest.kt
bystam: I will take a look!

bystam: Ah - the problem with that one is that the `InternalQuoteController` and the `ExternalQuoteController` have completely different APIs :/ when I build the "external one" with Areschougs help, he suggested moving to a newer model that he felt had a nicer API, so that's what we did. Therefore, reusing this isn't going to work really.

src/main/kotlin/com/hedvig/underwriter/web/ExternalQuoteController.kt
bystam: I kind of take that back. They do of course exist. They might just be re-created. My god I'm confused by what I have built myself haha

bystam: It's `connectToContracts` now :)

src/test/kotlin/com/hedvig/underwriter/testhelp/TestHttpClient.kt
bystam: 🙇

src/main/kotlin/com/hedvig/underwriter/web/ExternalQuoteController.kt
mightyost: :), I just looked in UW and guessed what was happening ;)...

Hmm, what is actually supposed to happen when calling endpoint?

mightyost: ...but maybe stupid to change this naming if consistent everywhere and people knows what it is...

src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/SelfChangeCommittingStrategy.kt
```diff
@@ -0,0 +1,37 @@
...
+        )
+        val changes = result.createdContracts + result.updatedContracts
+        changes.forEach { change ->
+            val quote = quotes.first { change.quoteId == it.id }
+            quoteRepository.update(
+                quote.copy(contractId = change.contractId, agreementId = change.agreementId, state = QuoteState.SIGNED)
+            )
+        }
+        return StartSignResponse.AlreadyCompleted
+    }
```
bystam, line 34: Sorry @mightyost - I definitely heard you when you said that you felt it was weird that a "signing strategy" immediately flushes data itself, and that this code should instead be moved out of this class as a special case of `AlreadyCompleted`.

What I found problematic with that approach when I started exploring it was:
- That means the concept of "self change" had to bleed into even more classes here
- It felt really weird to actually call `selfChangeContracts` inside `SignService` or somewhere else. `selfChangeContracts` is not some super-generic "write quotes to product-pricing", it's a bit more niche than that. Placing it outside of this specific signing strategy felt like it would risk bleeding outside of this very special case

bystam: Maybe we could aim to make a more generic version of `@PostMapping("/_/contracts/selfChange")` where it serves as a more general-purpose "change my contracts". But I don't think I feel that that's what it is right now...

src/main/kotlin/com/hedvig/underwriter/service/model/StartSignResponse.kt
```diff
@@ -22,6 +22,10 @@ sealed class StartSignResponse {
         val id: UUID
     ) : StartSignResponse()
 
+    object AlreadyCompleted : StartSignResponse() {
+        val ignore = true
```
mightyost, line 26: Hmm, looks strange. What is the `ignore` used for?

Why not a data class if it has vals?

src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/SelfChangeCommittingStrategy.kt
mightyost: Understand!

mightyost (approved):
Looks good.

src/main/kotlin/com/hedvig/underwriter/service/model/StartSignResponse.kt
bystam: It's because it's illegal to have an empty type in GraphQL, so the workaround is to give it some kind of dummy property.

PR #212, opened 2021-03-25
get rid of (unused) TeeFilter to avoid stack trace vomit

bystam:
# Jira Issue: [] 

## What?
- Remove `TeeFilter`, which gives functionality we don't really use 

## Why?
- `TeeFilter` has [a bug](https://jira.qos.ch/browse/LOGBACK-1527) in it that makes it print stack traces to stderr, which makes us unable to control its format. This causes Datadog to get massive piles of over-reported ERROR statements that drown out the real problems
- `TeeFilter` is used to retain body data for requests and responses so that can be printed. We don't actually use that data anyway, and they [even recommend you don't do that in production](http://logback.qos.ch/recipes/captureHttp.html#disabling)

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/config/LogBackAccess.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/config/LogBackAccess.kt
@@ -1,7 +1,6 @@
 package com.hedvig.underwriter.config
 
 import ch.qos.logback.access.tomcat.LogbackValve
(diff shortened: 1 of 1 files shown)


mightyost (approved):
Gr8!

meletios (approved):
🫖

PR #213, opened 2021-03-29
Block re-quoting

mightyost:
# Jira Issue: [IPC-67] 

## What?
Introduced blocking for customers to create quotes if already having an active agreement.

Address / birthday / ssn is used to identify if quote request is originating from a customer with an active agreement. 

All existing insurance types are supported except for Norwegian travel that has no address information or other info to make it possible to identify customer.

The blockings are registered with a metrics counter and is visible in DataDog.

**NOTE**: This feature will be deployed with a feature flag to initially not block any requests, but just to log identified re-quotings to make sure it is safe to do the blocking.

## Why?
Prices is going to be changed for SE on 1:st of April. To avoid unhappy customers (and headache for IEX) we are not going to serve quotes for customers already having an active agreement.

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [x] Tested locally

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
@@ -130,6 +130,90 @@ interface QuoteDao {
 
     @SqlQuery(
         """
+            SELECT DISTINCT
+            qr.master_quote_id
+            FROM quote_revision_apartment_data qrd
+            INNER JOIN quote_revisions qr
+            ON qrd.internal_id = qr.quote_apartment_data_id
+            WHERE 
+            qrd.street = :street
+            AND qrd.zip_code = :zipCode
+    """
+    )
+    fun findQuoteIdsBySwedishApartmentDataAddress(@Bind street: String, @Bind zipCode: String): List<UUID>
+
+    @SqlQuery(
+        """
+            SELECT DISTINCT
+            qr.master_quote_id
+            FROM quote_revision_house_data qrd
+            INNER JOIN quote_revisions qr
+            ON qrd.internal_id = qr.quote_house_data_id
+            WHERE 
+            qrd.street = :street
+            AND qrd.zip_code = :zipCode
+    """
+    )
+    fun findQuoteIdsBySwedishHouseDataAddress(@Bind street: String, @Bind zipCode: String): List<UUID>
+
+    @SqlQuery(
+        """
+            SELECT DISTINCT
+            qr.master_quote_id
+            FROM quote_revision_norwegian_home_contents_data qrd
+            INNER JOIN quote_revisions qr
+            ON qrd.internal_id = qr.quote_norwegian_home_contents_data_id
+            WHERE 
+            qrd.street = :street
+            AND qrd.zip_code = :zipCode
+    """
+    )
+    fun findQuoteIdsByNorwegianHomeContentsDataAddress(@Bind street: String, @Bind zipCode: String): List<UUID>
+
+    @SqlQuery(
+        """
+            SELECT DISTINCT
+            qr.master_quote_id
+            FROM quote_revision_danish_home_contents_data qrd
+            INNER JOIN quote_revisions qr
+            ON qrd.internal_id = qr.quote_danish_home_contents_data_id
+            WHERE 
+            qrd.street = :street
+            AND qrd.zip_code = :zipCode
+    """
+    )
+    fun findQuoteIdsByDanishHomeContentsDataAddress(@Bind street: String, @Bind zipCode: String): List<UUID>
+
+    @SqlQuery(
+        """
+            SELECT DISTINCT
+            qr.master_quote_id
+            FROM quote_revision_danish_accident_data qrd
+            INNER JOIN quote_revisions qr
+            ON qrd.internal_id = qr.quote_danish_accident_data_id
+            WHERE 
+            qrd.street = :street
+            AND qrd.zip_code = :zipCode
+    """
+    )
+    fun findQuoteIdsByDanishAccidentDataAddress(@Bind street: String, @Bind zipCode: String): List<UUID>
+
+    @SqlQuery(
+        """
+            SELECT DISTINCT
+            qr.master_quote_id
+            FROM quote_revision_danish_travel_data qrd
+            INNER JOIN quote_revisions qr
+            ON qrd.internal_id = qr.quote_danish_travel_data_id
+            WHERE 
+            qrd.street = :street
+            AND qrd.zip_code = :zipCode
+    """
+    )
+    fun findQuoteIdsByDanishTravelDataAddress(@Bind street: String, @Bind zipCode: String): List<UUID>
+
+    @SqlQuery(
+        """
             SELECT
             DISTINCT ON (qr.master_quote_id)
 
--- a/src/main/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImpl.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImpl.kt
@@ -5,8 +5,10 @@ import org.jdbi.v3.core.Jdbi
 import org.jdbi.v3.core.kotlin.inTransactionUnchecked
 import org.jdbi.v3.sqlobject.kotlin.attach
 import org.springframework.stereotype.Component
+import java.lang.IllegalArgumentException
 import java.time.Instant
 import java.util.UUID
+import kotlin.reflect.KClass
 
 @Component
 class QuoteRepositoryImpl(private val jdbi: Jdbi) : QuoteRepository {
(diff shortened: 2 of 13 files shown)


src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
```diff
@@ -128,6 +128,90 @@ interface QuoteDao {
...
+        """
+            SELECT DISTINCT
+            qr.master_quote_id
+            FROM quote_revision_apartment_data qrd
+            INNER JOIN quote_revisions qr
+            ON qrd.internal_id = qr.quote_apartment_data_id
+            WHERE 
+            qrd.street = :street
+            AND qrd.zip_code = :zipCode
+    """
```
Michael-Duivestein, line 141: Would it be possible to align the start and end quotes? (Same for all below)

src/main/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImpl.kt
```diff
@@ -146,6 +148,24 @@ class QuoteRepositoryImpl(private val jdbi: Jdbi) : QuoteRepository {
         }
     }
 
+    override fun findQuotesByAddress(street: String, zipCode: String, type: KClass<*>): List<Quote> = jdbi.inTransaction<List<Quote>, RuntimeException> { h ->
```
Michael-Duivestein, line 151: Maybe break this into multiple lines to make reading easier?
Also, maybe a more descriptive name for `h`?

src/main/kotlin/com/hedvig/underwriter/service/RequotingService.kt
```diff
@@ -0,0 +1,99 @@
...
+        }
+
+        logger.debug("Found ${quotes.size} existing quote(s) for same address: {}", quotes.joinToString(separator = ", ") { it.id.toString() })
+
+        // Filter out all signed quotes for user's ssn and birth date having an agreement
+        quotes = quotes
+            .filter { it.state == QuoteState.SIGNED }
+            .filter { it.agreementId != null }
+            .filter { it.birthDateMaybe == quote.birthDateMaybe }
+            .filter { quote.ssnMaybe == null || it.ssnMaybe == null || it.ssnMaybe == quote.ssnMaybe }
```
Michael-Duivestein, line 71: Are we getting all quotes for a given address, then filtering out any quotes that do not belong to the person getting the quote?
Will we also check for co-insured?

src/test/resources/application.yml
```diff
@@ -4,5 +4,11 @@ spring:
...
     allow-bean-definition-overriding: true
 
+  profiles:
+    active: development
+
 graphcms.project: Underwriter
 graphcms.url: empty
+
+features:
+  block-requoting: true
```
Michael-Duivestein, line 14: Newline

src/main/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImpl.kt
```diff
@@ -146,6 +148,24 @@ class QuoteRepositoryImpl(private val jdbi: Jdbi) : QuoteRepository {
...
+                NorwegianHomeContentsData::class -> dao.findQuoteIdsByNorwegianHomeContentsDataAddress(street, zipCode)
+                NorwegianTravelData::class -> emptyList()
+                DanishHomeContentsData::class -> dao.findQuoteIdsByDanishHomeContentsDataAddress(street, zipCode)
+                DanishAccidentData::class -> dao.findQuoteIdsByDanishAccidentDataAddress(street, zipCode)
+                DanishTravelData::class -> dao.findQuoteIdsByDanishTravelDataAddress(street, zipCode)
+                else -> throw IllegalArgumentException("Not supported")
+            }
+
+        findQuotes(quoteDatas, h)
+    }
```
FredrikAreschoug, line 167: If you would expect `type` to be `QuoteData` instead anyone that creates a new `QuoteData` type will have to handle that case and not blow up with an IllegalArgumentException. I agree that its a bit strange to send in the data clas since you are only interested in the type. But I did not come up with any better solution.

```suggestion
    override fun findQuotesByAddress(street: String, zipCode: String, type: QuoteData): List<Quote> = jdbi.inTransaction<List<Quote>, RuntimeException> { h ->
        val dao = h.attach<QuoteDao>()

        val quoteDatas =
            when (type) {
                is SwedishHouseData -> dao.findQuoteIdsBySwedishApartmentDataAddress(street, zipCode)
                is SwedishHouseData -> dao.findQuoteIdsBySwedishHouseDataAddress(street, zipCode)
                is NorwegianHomeContentsData -> dao.findQuoteIdsByNorwegianHomeContentsDataAddress(street, zipCode)
                is NorwegianTravelData -> emptyList()
                is DanishHomeContentsData -> dao.findQuoteIdsByDanishHomeContentsDataAddress(street, zipCode)
                is DanishAccidentData -> dao.findQuoteIdsByDanishAccidentDataAddress(street, zipCode)
                is DanishTravelData -> dao.findQuoteIdsByDanishTravelDataAddress(street, zipCode)
            }

        findQuotes(quoteDatas, h)
    }
```

vonElfvin:
Do we really know members will come in and check their price and give IEX a headache? Can't we try and see before we deploy something like this? 

This would make the API less predictable and give a weird user experience. Honestly I think not being able to create a quote would give more headache than showing a different price. Thus, this really feels like suboptimizing for me 🤔 

If I am totally off here please enlighten me!! 🙏 😇

src/main/kotlin/com/hedvig/underwriter/service/UnderwriterImpl.kt
```diff
@@ -31,9 +38,14 @@ import javax.money.MonetaryAmount
...
     private val priceEngineService: PriceEngineService,
     private val quoteStrategyService: QuoteStrategyService,
-    private val metrics: Metrics
+    private val requotingService: RequotingService,
+    private val blockedByTypeCounter: BlockedByTypeCounter,
+    private val breachedGuidelinesCounter: BreachedGuidelinesCounter
 ) : Underwriter {
 
+    @Value("\${features.block-requoting}")
+    private var blockRequoting: Boolean = false
```
FredrikAreschoug, line 47: Just a heads up I think this will always be false even if you add `features.block-requoting: true` to the config map.

I think this will do the trick:
```suggestion
    @Value("\${features.block-requoting:false}")
    lateinit var blockRequoting: Boolean
```

FredrikAreschoug (commented):
Looks good added som small comments

src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
mightyost: Fixed!

You are like me :), things should be tidy...

src/main/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImpl.kt
mightyost: Did align it with how the other functions do it... I did not change the `h`...

src/main/kotlin/com/hedvig/underwriter/service/RequotingService.kt
mightyost: Yes... Its a bit probabilistic...

We are not checking co-insured or any other fields than address, birthday and ssn...

We will have metrics in place and monitor this. If things goes out of hand or can be approved we can do it along the way...

src/test/resources/application.yml
mightyost: Done

mightyost:
> Do we really know members will come in and check their price and give IEX a headache? Can't we try and see before we deploy something like this?
> 
> This would make the API less predictable and give a weird user experience. Honestly I think not being able to create a quote would give more headache than showing a different price. Thus, this really feels like suboptimizing for me 🤔
> 
> If I am totally off here please enlighten me!! 🙏 😇

Yes, I guess @meletios can fill you in on the details and extent of this. According to my discussions with Anton this feature is implemented by all insurance companies and is very wanted here as well...

src/main/kotlin/com/hedvig/underwriter/model/QuoteRepositoryImpl.kt
mightyost: Yes I was thinking along the same lines... Migrated to your suggestion!

src/main/kotlin/com/hedvig/underwriter/service/UnderwriterImpl.kt
mightyost: `lateinit` on primitives does not work. But thanks for the tip about setting a default value in the `@Value` annotation. I am using that!

FredrikAreschoug: yes forgot about that :blush:

FredrikAreschoug (approved):
Nice! :ice_cream:

PR #215, opened 2021-03-31
Changed GITHUB_USER key to match right env var in Codefresh.

mightyost:
## What?
- Changed GITHUB_USER key to match right env var in Codefresh.

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,10 +2,10 @@
 FROM maven:3.6.3-amazoncorretto-11 AS dependencies
 WORKDIR /usr/app
 
-ARG GITHUB_USERNAME
+ARG GITHUB_USER
 ARG GITHUB_TOKEN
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -2,10 +2,10 @@
 FROM maven:3.6.3-amazoncorretto-11 AS dependencies
 WORKDIR /usr/app
 
-ARG GITHUB_USERNAME
```
Michael-Duivestein, line 5: Isn't this used elsewhere as well? If that's the case would it be better to change codefresh instead?

fredriklagerblad (approved):

Dockerfile
mightyost: It is only used in Rapio and in Underwriter. Rapio uses the right arg while underwriter does not...

PR #218, opened 2021-04-06
Enabled blocking of re-quoting in prod.

mightyost:
# Jira Issue: [IPC-67] 

## What?
- Block re-quoting in prod if already having an active agreement.

[IPC-67]: https://hedvig.atlassian.net/browse/IPC-67


meletios (approved):
🚀

PR #220, opened 2021-04-06
Enable requoting in staging

meletios:
# Jira Issue: [] 

## What?
- Enable requoting in staging

## Why?
- It will help testing in staging for the webbies!! 😍🥰

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -52,6 +52,10 @@ lokalise:
 hedvig:
   notification-service.url: http://notification-service
 
+---
+spring:
+  profiles: production
+
 features:
   block-requoting: true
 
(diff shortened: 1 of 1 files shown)


AlexSJ1990 (approved):
✅

src/main/resources/application.yml
```diff
@@ -63,6 +63,9 @@ graphql:
   servlet:
     exception-handlers-enabled: true
 
+features:
+  block-requoting: false
+
 ---
```
mightyost, line 69: Hmm, I do not really know how this will work and in which order Spring handle profile config.

Since blocking is turned on in `production, staging` above and then turned of for `staging` here it feels a bit uncertain...

Maybe add a new section:
```
---
spring:
  profiles: production

 features:
   block-requoting: true

 ---
 ```
... and then remove the:
```
features:
  block-requoting: true
```
under `production, staging`.

?

mightyost (approved):
Gr8!

PR #221, opened 2021-04-07
Disable blocking creating quotes for customers with active an agreement in staging.

mightyost:
Disable the feature to block creating quotes for customers already having a signed agreement.

This since it makes it cumbersome to do testing in stage.

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -52,6 +52,10 @@ lokalise:
 hedvig:
   notification-service.url: http://notification-service
 
+---
+spring:
+  profiles: production
+
 features:
   block-requoting: true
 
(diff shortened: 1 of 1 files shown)


Michael-Duivestein (commented):
I think this is a duplicate of https://github.com/HedvigInsurance/underwriter/pull/221/files

mightyost:
> I think this is a duplicate of https://github.com/HedvigInsurance/underwriter/pull/221/files

@Michael-Duivestein, do you mean https://github.com/HedvigInsurance/underwriter/pull/220 ?

If so it is :). The difference is that the config in this PR is less ambiguous imho according to my last comment in PR 220...

meletios (approved):
Approved, even though its the same with #220 😅

PR #225, opened 2021-04-14
Migrate away from Bintray

mightyost:
# Jira Issue: [IPC-58] 

## What?
- 

## Why?
- 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

[IPC-58]: https://hedvig.atlassian.net/browse/IPC-58

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -29,14 +29,6 @@
 
     <repositories>
         <repository>
-            <snapshots>
-                <enabled>false</enabled>
-            </snapshots>
-            <id>bintray-hedvig-hedvig-java</id>
-            <name>bintray</name>
-            <url>https://dl.bintray.com/hedvig/hedvig-java</url>
-        </repository>
-        <repository>
             <id>github</id>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -250,7 +242,7 @@
         <dependency>
             <groupId>com.hedvig.graphql.commons</groupId>
             <artifactId>graphql-commons</artifactId>
-            <version>1.0.7</version>
+            <version>1.0.8</version>
```
Michael-Duivestein, line 245: Is 1.0.8 not available on Github?

Michael-Duivestein (approved):

pom.xml
mightyost: 1.0.7 = 1.0.8 and 1.0.8 only exist on GitHub. I bumped the version to be able to check downloads where working ok from GitHub while the other libs where still grabbed from bintray...

PR #227, opened 2021-04-23
Gdpr PII rollout

mightyost:
# Jira Issue: [IPC-72] 

## What?
- New endpoint for k8s cron job to detect and anonymise old quotes without any attached agreement. If any members for those quotes (that has no other quotes) will be deleted in Member Service, CustomerIO and API Gateway.

- Removed direct access to CustomerIO that is not used.

- Added new `updatedAt` field to `Quote` in order to see when last revision was created. The existing `createdAt` shows when the first revision was added.

Implementation Proposal:
https://docs.google.com/document/d/1tOOwFlDYwrZkQpogrq2Xlr4WN34QAA90D3yExSYPar4/edit#

## Why?
- GDPR

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [x] Tested locally

[IPC-72]: https://hedvig.atlassian.net/browse/IPC-72


src/main/resources/db/changelog/migrations/20210422-1030-add-deleted-quotes-table.sql
```diff
@@ -0,0 +1,15 @@
+--liquibase formatted.sql
+
+--changeset ostenforshed:20210422-1030-add-deleted-quotes-table.sql
+
+CREATE TABLE deleted_quotes (
+    quote_id uuid PRIMARY KEY,
+    created_at timestamp NOT NULL,
```
fredriklagerblad, line 7: Maybe use timestamptz instead ?

src/main/kotlin/com/hedvig/underwriter/service/GdprServiceImpl.kt
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    private fun hasNoOtherQuotes(memberId: String, quotes: Set<UUID>): Boolean =
+        quoteRepository.findByMemberId(memberId)
+            .map { it.id }
+            .all { quotes.contains(it) }
+
+    private fun deleteMembers(memberIds: List<String>) {
+        // Member Service
+        // Endpoint not available yet
```
fredriklagerblad, line 88: Guess the idea to wait for the endpoints to be ready, before running this live ?

src/main/kotlin/com/hedvig/underwriter/web/GdprController.kt
```diff
@@ -0,0 +1,24 @@
...
+import org.springframework.web.bind.annotation.RestController
+
+@RestController
+@RequestMapping("/_/v1/gdpr")
+class GdprController(
+    val gdprService: GdprService
+) {
+    @PostMapping("/clean")
+    @LogCall
+    fun clean(@RequestParam("dry-run") dryRun: Boolean?): ResponseEntity<out Any> {
```
fredriklagerblad, line 18: Will the other services's endpoints also support the dry-run param? So we can make a system-wide dry-run ...

fredriklagerblad (approved):
Nice! Just a few questions...

"Data Retentention" all over again :-)

src/main/resources/db/changelog/migrations/20210422-1030-add-deleted-quotes-table.sql
mightyost: Right or not? I chosen to align it with how the other timestamp columns are defined in Underwriter currently.

src/main/kotlin/com/hedvig/underwriter/service/GdprServiceImpl.kt
mightyost: Yes, it will be feature googled of in prod until these endpoints are in place.

src/main/resources/db/changelog/migrations/20210422-1030-add-deleted-quotes-table.sql
fredriklagerblad: OK.

src/main/kotlin/com/hedvig/underwriter/web/GdprController.kt
mightyost: Would be nice ;)

PR #234, opened 2021-05-03
add address details for danish accident and travel and send to price …

AlexSJ1990:
# Jira Issue: [IPC-110] 

## What?
- add address data to danish accident and travel quotes
- stop stubbing price-engine for danish accident and travel
- add tests

## Why?
- we will launch with accident and home content

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [ ] Tested locally

[IPC-110]: https://hedvig.atlassian.net/browse/IPC-110


src/main/kotlin/com/hedvig/underwriter/graphql/type/QuoteMapper.kt
```diff
@@ -205,12 +205,20 @@ class QuoteMapper(
...
+            city = quote.data.city,
+            apartment = quote.data.apartment,
+            floor = quote.data.floor,
             numberCoInsured = quote.data.coInsured,
             isStudent = quote.data.isStudent
         )
         is DanishTravelData -> QuoteSchema.DanishTravel(
             street = quote.data.street,
             zipCode = quote.data.zipCode,
+            bbrId = quote.data.bbrId,
```
vonElfvin, line 218: Does this mean that IEX needs to find the bbrId?

AlexSJ1990: nope - it will be empty for now, but when we implement autocomplete in hope we will populate it.

src/main/kotlin/com/hedvig/underwriter/service/UnderwriterImpl.kt
```diff
@@ -141,8 +141,9 @@ class UnderwriterImpl(
                 ).price
             }
             is DanishAccidentData -> {
-                // TODO: fix when pricing is in place
-                Money.of(6666, "DKK")
+                priceEngineService.queryDanishAccidentPrice(
+                    PriceQueryRequest.DanishAccident.from(quote.id, quote.memberId, quote.data)
+                ).price
             }
             is DanishTravelData -> {
```
fredriklagerblad, line 148: Maybe we should enable DanishTravel now as well, since you also added support for that in price-engine. While not provisioned in price-engine, any request with it should return an error, so should be "safe".

fredriklagerblad (approved):
Nice. Just a comment aboyt if we shouldn't also enabled DanishTravel all the way also.

PR #235, opened 2021-05-05
Support in GdprService for deleting members in API GW and Member Service

mightyost:
Added support in `GdprService` for deleting members in API GW and Member Service.


fredriklagerblad (approved):
Nice!

Michael-Duivestein (approved):

PR #247, opened 2021-05-20
Read up large about of quotes in chunks

mightyost:
We got error in prod where we cannot read up a large amount of quotes in one go.

Fixed so anonymised street adress does not have any json string incompatible chars breaking inserts into `deleted_quotes` table.

And som linting fixes...


fredriklagerblad (approved):
👍

PR #248, opened 2021-05-20
Added missing fk indexes

mightyost:
Discovered that we do not have any indexes on our foreign key columns. This is something that is automatically created in MySQL but not in Postgres:

https://www.cybertec-postgresql.com/en/index-your-foreign-key/

This will hopefully speed up some queries and deletes a bit.


Michael-Duivestein (approved):
That's a lot of indexes...

PR #251, opened 2021-05-25
Added an extra PP contract check to GDPR cleaning job.

mightyost:
There seems to be members created and signed before Underwriter was born. To not delete these we have added a new extra check to PP for the members to delete.

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [x] Tested locally


fredriklagerblad (approved):
:+1

PR #253, opened 2021-05-26
Feature/self change quote sign triaging

bystam:
# Jira Issue: [MX-148], [MX-21]

## What?
- Tag quotes with a new `QuoteInitiatedFrom` type (SELF_CHANGE)
- Use the new `initatedFrom` value to triage differently in `signMethodForQuotes`

## Why?
- The offer page needs to know that it should "sign" these quotes differently from any of the quotes that are produced in the onboarding flow
- This will help the apps target a different mutation when finalizing the moving flow (sadly this mutation is in Giraffe, not underwriter 😞)
- slack thread: https://hedviginsurance.slack.com/archives/C01NB6GBRFS/p1622013719001900

## Help
- I am not sure about this at all. Adding a new `QuoteInitiatedFrom` case feels somewhat uncontroversial (even if it somehow feels weird that IOS/ANDROID are cases on the same axis as SELF_CHANGE). But actually triaging this to a `SignMethod` feels like a real hack. This means I have introduced a sign method that actually doesn't exist. It doesn't match the signing strategies, and will become very confusing going forward.
- I feel that `SignMethod`/`SignStrategy` and friends seem very coupled to the specific flow of **signing** quotes as part of the onboarding. This feels like a proper case of shoehorning something into where it really shouldn't be.

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [ ] Tested locally

[MX-148]: https://hedvig.atlassian.net/browse/MX-148
[MX-21]: https://hedvig.atlassian.net/browse/MX-21


src/main/kotlin/com/hedvig/underwriter/model/QuoteInitiatedFrom.kt
```diff
@@ -6,5 +6,10 @@ enum class QuoteInitiatedFrom {
     APP,
     IOS,
     ANDROID,
-    HOPE
+    HOPE,
+
+    /**
+     * The quote was created as part of the "self change" flow (moving flow) in the apps.
+     */
+    SELF_CHANGE
```
meletios, line 14: Nice! Is the new enum type going to affect blocking/re-quoting Mr @mightyost ?

bystam: I personally have no idea

PR #254, opened 2021-05-26
Fixed `priceFrom` on quote.

mightyost:
The `priceFrom` should point at the last quote before price was changed and not the last "same" quote as before.

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [ ] Tested locally


fredriklagerblad (approved):
👍

PR #255, opened 2021-05-27
Refactored integration tests.

mightyost:
## What?
Added a new class `IntegrationTest` with mocks for all clients. This class can be extended for integration tests to reuse the same spring boot application and in that way save execution time.

## Why?
To cut down execution time from 50s to 25s for current 4 integration tests.

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

### DIFF
--- a/src/test/kotlin/com/hedvig/underwriter/testhelp/IntegrationTest.kt
+++ b/src/test/kotlin/com/hedvig/underwriter/testhelp/IntegrationTest.kt
@@ -6,11 +6,8 @@ import com.hedvig.underwriter.serviceIntegration.notificationService.Notificatio
 import com.hedvig.underwriter.serviceIntegration.priceEngine.PriceEngineClient
 import com.hedvig.underwriter.serviceIntegration.productPricing.ProductPricingClient
 import com.ninjasquad.springmockk.MockkBean
-import org.junit.runner.RunWith
 import org.springframework.boot.test.context.SpringBootTest
-import org.springframework.test.context.junit4.SpringRunner
 
-@RunWith(SpringRunner::class)
 @SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
 class IntegrationTest {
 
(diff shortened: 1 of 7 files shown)


src/test/kotlin/com/hedvig/underwriter/testhelp/IntegrationTest.kt
```diff
@@ -0,0 +1,31 @@
...
+import com.hedvig.underwriter.serviceIntegration.memberService.MemberServiceClient
+import com.hedvig.underwriter.serviceIntegration.notificationService.NotificationServiceClient
+import com.hedvig.underwriter.serviceIntegration.priceEngine.PriceEngineClient
+import com.hedvig.underwriter.serviceIntegration.productPricing.ProductPricingClient
+import com.ninjasquad.springmockk.MockkBean
+import org.junit.runner.RunWith
+import org.springframework.boot.test.context.SpringBootTest
+import org.springframework.test.context.junit4.SpringRunner
+
+@RunWith(SpringRunner::class)
```
bystam, line 13: `SpringBootTest` already has `@ExtendWith(SpringExtension::class)`. I think also adding `RunWith` will make it run half JUnit4 half JUnit5 which might cause weird behaviour. I suggest just dropping `@RunWith(SpringRunner::class)` entirely and just say that people who extend this class must use JUnit5

bystam (approved):
This looks cool! I suggest dropping `@RunWith` on the base class and refactoring the tests using it to use JUnit 5 instead (should just be a quite simple import change and renaming of `@Before` -> `@BeforeEach`)

src/test/kotlin/com/hedvig/underwriter/testhelp/IntegrationTest.kt
Michael-Duivestein: See this PR for inspiration:
https://github.com/HedvigInsurance/product-pricing/pull/756

PR #258, opened 2021-06-02
Feature/get address data from street

AlexSJ1990:
# Jira Issue: [] 

## What?
- 

## Why?
- 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

### DIFF
--- a/src/test/resources/mutations/danishStreetToAddress/createDanishHomeContentsQuoteWithApartmentAndStreetNotNull.graphql
+++ b/src/test/resources/mutations/danishStreetToAddress/createDanishHomeContentsQuoteWithApartmentAndStreetNotNull.graphql
@@ -6,10 +6,10 @@ mutation {
         birthDate: "1961-12-21"
         email: "test@email.com"
         danishHomeContents: {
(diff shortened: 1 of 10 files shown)


src/test/resources/mutations/danishStreetToAddress/createDanishHomeContentsQuoteWithApartmentAndStreetNotNull.graphql
```diff
@@ -0,0 +1,60 @@
...
+                    bbrId
+                    livingSpace
+                    coInsured
+                    isStudent
+                    danishHomeContentType: type
+                }
+            }
+        }
+    }
+}
```
Michael-Duivestein, line 60: newlines

Michael-Duivestein (approved):
After chatting with you I understand this is a "temporary" solution until we get autocomplete for Danish addresses. 
I still think this is patching a flaw in our system rather than fixing it. A better solution feels like using multiple fields for a Danish address until autocomplete is done.

PR #261, opened 2021-06-09
Do not do blocking and price reuse check for Leadvägen 1

mightyost:
# Jira Issue: [IPC-149] 

## What?
We have a client creating quotes for this street every 5 min resulting in 10k+ quotes to fetch from db every time.

## Why?
To protect the db and system.

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [x] Tested locally

[IPC-149]: https://hedvig.atlassian.net/browse/IPC-149

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/service/RequotingServiceImpl.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/service/RequotingServiceImpl.kt
@@ -49,6 +49,12 @@ class RequotingServiceImpl(
                 return false
             }
 
+            // Do not process certain "expensive" quotes
+            if (isTooExpensive(quote)) {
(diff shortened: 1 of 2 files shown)


vonElfvin (approved):
Who would do such a thing

Michael-Duivestein (approved):
Looks good! If someone we're integrating with is using the to check we're up:
1) Will their status check still work if we make this change?
2) We need to provide them with a better way to do this check.

mightyost:
> Looks good! If someone we're integrating with is using the to check we're up:
> 
> 1. Will their status check still work if we make this change?
> 2. We need to provide them with a better way to do this check.

1. Yes, maybe working better since we will not block some of then ;)
2. Yes, this is not optimal! But do not really know the reason for these requests thou...

PR #262, opened 2021-06-15
MX-181: enable signing quotes with memberId injected, making underwriter not create a new member

bystam:
# Jira Issue: [MX-181] 

## What?
- Enable injecting memberId when signing from Rapio, making it sign with that member rather than creating a new one

## Why?
- Members will be created before signing in the new AVY partnership use case.

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

[MX-181]: https://hedvig.atlassian.net/browse/MX-181

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/web/dtos/SignQuotesRequestDto.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/web/dtos/SignQuotesRequestDto.kt
@@ -13,17 +13,10 @@ data class SignQuotesRequestDto(
     val startDate: LocalDate?,
     val insuranceCompany: String?,
     @Masked val email: String,
-    val price: BigDecimal? = null, // Used for bundle verification
-    val currency: String? = null
-) {
-    companion object {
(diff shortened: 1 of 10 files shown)


src/main/kotlin/com/hedvig/underwriter/web/dtos/SignQuotesRequestDto.kt
```diff
@@ -13,17 +13,10 @@ data class SignQuotesRequestDto(
...
-        fun from(quoteId: UUID, request: SignQuoteRequestDto): SignQuotesRequestDto = SignQuotesRequestDto(
-            listOf(quoteId),
-            name = request.name,
-            ssn = request.ssn,
-            startDate = request.startDate,
-            insuranceCompany = request.insuranceCompany,
-            email = request.email
-        )
-    }
-}
```
bystam, line 29: Unused 🤷

fredriklagerblad (approved):
Looks nice!
There is a `RapioIntegrationTest` class, if you add add tests of this specific functionality.

bystam:
> RapioIntegrationTest



> Looks nice!
> There is a `RapioIntegrationTest` class, if you add add tests of this specific functionality.

Updated with a test :)

PR #263, opened 2021-06-17
MX-181: validate that injected members when signing from rapio don't already have a contract

bystam:
# Jira Issue: [MX-181] 

## What?
- When a "sign quotes from rapio" request comes in with an explicitly injected member ID, verify that said member don't already have a contract.

## Why?
- We want to enable signing of inejcted member IDs in the following cases:
  - the member only has a trial right now
  - the member does not have a trial, not a contract - i.e. it is "free"

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

[MX-181]: https://hedvig.atlassian.net/browse/MX-181

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
@@ -255,7 +255,7 @@ class SignServiceImpl(
         validateQuotesToSignFromRapio(quotes)
         validateBundlePrice(quotes, request.price, request.currency)
 
-        val memberId = request.memberId ?: getCreateMember(quotes)
+        val memberId = validateOrCreateMember(request.memberId, quotes)
 
         quotes = quotes
             .map { it.copy(memberId = memberId) }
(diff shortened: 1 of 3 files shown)


src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
```diff
@@ -341,41 +341,40 @@ class SignServiceImpl(
         }
     }
 
-    private fun getCreateMember(quotes: List<Quote>): String {
+    private fun validateOrCreateMember(explicitMemberId: String?, quotes: List<Quote>): String {
```
bystam, line 344: This is not a super-thought through way of protecting the integrity of our state. We are mostly doing what we can to try and make sure our member state isn't corrupted. But this is probably really scary and/or error prone.

src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
```diff
@@ -341,41 +341,40 @@ class SignServiceImpl(
...
         // Does this member already exist?
         val memberAlreadySigned = memberService.isSsnAlreadySignedMemberEntity(ssn).ssnAlreadySignedMember
 
         if (memberAlreadySigned) {
             throw ErrorException(ErrorCodes.MEMBER_HAS_EXISTING_INSURANCE, "Member already have signed quote(s)")
         }
 
-        memberId = memberService.createMember()
-        logger.info("New member created: $memberId")
+        val created = memberService.createMember()
```
vonElfvin, line 363: Maybe?
```suggestion
        val createdMemberId = memberService.createMember()
```

vonElfvin (approved):
Neat!

PR #264, opened 2021-06-21
Ensure member has contract before creating contract from hope

vonElfvin:
# Jira Issue: [MX-185] 

## What?
- Ensure member has contract before creating contract from hope

## Why?
- So we don't create contracts from Hope for members with Trials

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

[MX-185]: https://hedvig.atlassian.net/browse/MX-185


bystam (approved):
Makes sense to me.

PR #268, opened 2021-06-30
Use Insurely net price

fredriklagerblad:
# Jira Issue: [IPC-170] 

## What?
- Use the net competitor price from Insurely instead of gross, when matching prices.

## Why?
- To offer more competetive prices 
- And cause the pricing guys says so :-) 

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [ ] Tested locally

[IPC-170]: https://hedvig.atlassian.net/browse/IPC-170


meletios (approved):
✅

PR #269, opened 2021-07-01
Feat/override quote price

fredriklagerblad:
# Jira Issue: [IPC-134] 

## What?
- Add overrideQuotePrice REST endpoint, which allows for overriding quote's price
- Input parameters: quoteId, price and overiddenBy - will be called by backoffice
- UW basically sets those params and performs a validateAndCompleteQuote
- The overridden price and overiddenBy are forwarded to price-engine, so it can use them to calculate the correct final price and line items (e.g. tax), and also audit-log the details
- We explicitly avoid reusing any old price for these requests

## Why?
- To allow Hope users to override a quote's price

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [x] Tested locally

[IPC-134]: https://hedvig.atlassian.net/browse/IPC-134

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
@@ -37,7 +37,7 @@ import com.hedvig.underwriter.web.dtos.ErrorCodes
 import com.hedvig.underwriter.web.dtos.ErrorResponseDto
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.stereotype.Service
-import java.lang.IllegalStateException
+import java.math.BigDecimal
 import java.security.MessageDigest
 import java.time.Instant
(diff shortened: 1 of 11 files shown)


src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -86,6 +86,36 @@ class QuoteServiceImpl(
...
     }
 
+    override fun overrideQuotePrice(
+        quoteId: UUID,
+        price: BigDecimal,
+        overriddenBy: String
+    ): Either<ErrorResponseDto, Quote> {
+
+        return findQuoteOrError(quoteId)
+            .filterOrOther(
```
Michael-Duivestein, line 96: Didn't know this existed. So useful!

Michael-Duivestein (approved):
Looks good to me, but would recommend someone else also approve.

PR #271, opened 2021-07-06
Fix embedded postgres on m1 macs

tobiasbexelius:
# Jira Issue: [] 

## What?
- Swap out embedded postgres to this [fork](https://github.com/zonkyio/embedded-postgres), which supports m1 macs without any hassle

## Why?
- So that people with m1 macs can run tests that rely on embedded postgres

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -42,10 +42,10 @@
             <version>1.0.1</version>
         </dependency>
         <dependency>
-           <groupId>com.hedvig.libs</groupId>
-           <artifactId>translations</artifactId>
-           <version>0.3.0</version>
-       </dependency>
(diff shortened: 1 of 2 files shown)


bystam (approved):
Oof! I wonder if this also means that we no longer need to have postgres installed locally to run the tests

tobiasbexelius:
> Oof! I wonder if this also means that we no longer need to have postgres installed locally to run the tests

How do you mean? I don't think this fork changes anything fundamental about embedded postgres, but maybe I missed something

PR #272, opened 2021-07-06
MX-196: Create custom approveQuotes mutation instead of piggy-backing on the existing signQuotes

bystam:
# Jira Issue: [MX-196] 

## What?
- Create a new codepath for finalizing selfChange quotes (moving flow)

## Why?
- The existing `signQuotes(...)` was too tailored to the onboarding use case, which means it couldn't be run at all in practice
- Specifically, it required the member to be not signed since before, which is 100% the opposite of the moving flow use case

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

[MX-196]: https://hedvig.atlassian.net/browse/MX-196


src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
```diff
@@ -422,12 +423,21 @@ class SignServiceImpl(
...
+        require(quotes.all { it.initiatedFrom == QuoteInitiatedFrom.SELF_CHANGE }) {
+            "Can only approve self-change quotes"
+        }
+        val result = productPricingService.selfChangeContracts(
+            memberId = memberId,
+            quotes = quotes
+        )
+        val changes = result.createdContracts + result.updatedContracts
+        changes.forEach { change ->
+            val quote = quotes.first { change.quoteId == it.id }
```
vonElfvin, line 437: This just made me think about that there should be a operation called like `innerJoin` and `leftJoin` etc. that works exactly like SQL joins and creates pairs of matches, on this case by ID 🤔

src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/SignStrategyService.kt
```diff
@@ -113,6 +110,12 @@ class SignStrategyService(
             this.any { it.data is DanishTravelData }
 
     override fun getSignMethod(quotes: List<Quote>): SignMethod {
+        val isSelfChangeQuotes = quotes.any { it.initiatedFrom == QuoteInitiatedFrom.SELF_CHANGE }
```
vonElfvin, line 113: Insure how to do this. Guess we need to validate the opposite as well 🤔 
```suggestion
        val isSelfChangeQuotes = quotes.all { it.initiatedFrom == QuoteInitiatedFrom.SELF_CHANGE }
```

vonElfvin: Or maybe use `any` and break further down if it was a mixed bag

vonElfvin (approved):
Great job! 👍

tobiasbexelius (approved):
Nice!

src/main/kotlin/com/hedvig/underwriter/service/quotesSignDataStrategies/SignStrategyService.kt
bystam: Well, I think the root cause is the model being a "list of quotes" and not a single quote. They need to have certain properties perfectly aligned.

The mega-defensive solution would be to crash if the `initiatedFrom` value is not the same.

But my half-defensive model here is to say that "if one of them is SELF_CHANGE", then `approveOnly` is the sign method. Then in the sign method itself, we fail if they are not all `SELF_CHANGE`.

bystam: I don't find that any of them are obviously better honestly

vonElfvin: Yea signing is weird. I say go with this

src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
tobiasbexelius: Something like this @vonElfvin ?

```
fun <A, B, K, R> Iterable<A>.join(
    other: Iterable<B>,
    selector: (A) -> K,
    otherSelector: (B) -> K,
    action: (A, B) -> R
) {
    forEach { a ->
        val ak = selector(a)
        other.forEach { b ->
            val bk = otherSelector(b)
            if (ak == bk) {
                action(a, b)
            }
        }
    }
}
```

PR #273, opened 2021-07-06
dont set bbrId from hope

AlexSJ1990:
# Jira Issue: [] 

## What?
- from hope we always send in all quote data when updating a quote. This means that if a member has gone through the onboarding with autocomplete and then iex amend the address details the old brrId from the previous quote will get sent (unless iex delete the bbrId field).
- In underwriter, we expect that when we edit a quote we will only get data for fields that will be updated. It turns out that from hope and the web this is not the case (though this is the case for apps). We will need to align and fix this (but that will come in another pr).
- this removes bbrId from the quote schema as we are not using autocomplete in hope. This means the bbrId will be set to null in all cases.
- If a member has changed something (other than the address) on the quote, geomatic should still be able to find them based on the other address fields even though the brrId is null.

## Why?
- 

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally

### DIFF
--- a/src/main/kotlin/com/hedvig/underwriter/graphql/type/QuoteMapper.kt
+++ b/src/main/kotlin/com/hedvig/underwriter/graphql/type/QuoteMapper.kt
@@ -194,7 +194,6 @@ class QuoteMapper(
         is DanishHomeContentsData -> QuoteSchema.DanishHomeContent(
             street = quote.data.street,
             zipCode = quote.data.zipCode,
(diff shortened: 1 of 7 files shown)


fredriklagerblad (approved):
I think the reasoning is valid, and it should work.
Maybe it would also be good with (integration?) test that tests the QuoteSchemaController.

PR #274, opened 2021-07-08
Add a test to assert uw breach code

tobiasbexelius:
# Jira Issue: [] 

## What?
-  A new test and some changed formatting

## Why?
- Part of a bug hunt. The bug turned out to be elsewhere, but this might still be nice to have

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [x] Tested locally


src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
```diff
@@ -333,9 +333,14 @@ class QuoteServiceImpl(
...
-                    "quote cannot be calculated, underwriting guidelines are breached [Quote: $quote",
-                    quote.breachedUnderwritingGuidelines.map { BreachedGuideline("Deprecated", it) }
+                    errorCode = ErrorCodes.MEMBER_BREACHES_UW_GUIDELINES,
+                    errorMessage = "quote cannot be calculated, underwriting guidelines are breached [Quote: $quote]",
+                    breachedUnderwritingGuidelines = quote.breachedUnderwritingGuidelines.map {
+                        BreachedGuideline(
+                            message = "Deprecated",
+                            code = it
+                        )
+                    }
```
vonElfvin, line 343: Can't this list just be empty? 🤔

vonElfvin (approved):
👍

src/main/kotlin/com/hedvig/underwriter/service/QuoteServiceImpl.kt
tobiasbexelius: It will be empty if there are no breached guidelines on the quote, or how do you mean? :)

PR #275, opened 2021-07-08
Add local service urls

tobiasbexelius:
# Jira Issue: [] 

## What?
- Add local service urls

## Why?
- So the service can be run locally without an application-development file

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


bystam (approved):
Great! That was missing :D

PR #280, opened 2021-08-20
feat/lookup-service-client-404-null

fredriklagerblad:
# Jira Issue: [] 

## What?
- Make lookupService's feign client return null for 404, instead of exception

## Why?
- To not trigger an ERROR for these cases, only a warning

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [x] Tested locally


mightyost (approved):
👍

AlexSJ1990 (approved):

PR #287, opened 2021-08-31
Turn off all the Spring-based Liquibase, as we are running it manually and they are conflicting

bystam:
# Jira Issue: [] 

## What?
- Never run Liquibase using spring auto-configuration
- Always run Liquibase on start unless explicitly disabled

## Why?
- We want to run Liquibase by default, but never through spring auto conf

## Optional checklist
- [x] Codescouted
- [ ] Unit tests written
- [x] Tested locally


tobiasbexelius (approved):
🎉

PR #290, opened 2021-09-05
POC for introducing generic promise stacks / products.

mightyost:
Note that this is just a test to verify the ideas around the new generic flow, it is here for pure illustration and to spark discussions for the way forward.

Generic Swedish Accident is added as a test and it is possible to create a quote and sign it.

See `RapioIntegrationTest.'Create Generic quote and sign it successfully'()` for how to create a quote and sign it.

Not currently covered:

- Update quote requests
- Bundling
- Signing strategies
- HOPE Schemas
- Guidelines


src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
```diff
@@ -746,3 +789,40 @@ interface QuoteDao {
...
     fun deleteLineItems(@Bind revisionId: Int)
 }
+
+internal class GenericDataMapper(val objectMapper: ObjectMapper) : RowMapper<GenericData?> {
+
+    override fun map(r: ResultSet, ctx: StatementContext?): GenericData {
+
+        val internalId = r.getInt("internal_id")
+        val id = r.getObject("id", UUID::class.java)
+        val market = r.getString("market")
```
bystam, line 799: Can't we just save the country code here?

src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
```diff
@@ -746,3 +789,40 @@ interface QuoteDao {
...
+
+        val product = objectMapper.readValue(productJson, Products.getClass(productType).java)
+
+        return GenericData(
+            id,
+            CountryCode.valueOf(market),
+            HolderPerson(firstName, lastName, ssn, birthdate?.toLocalDate(), email, phone),
+            productType,
+            product,
+            internalId
```
bystam, line 817: Will this have to be updated for every new kind of promise stack? Or is that what the `Products` thing is for? Maybe these are all the generic values of any promise stack?

src/main/kotlin/com/hedvig/underwriter/products/model/Product.kt
```diff
@@ -0,0 +1,14 @@
...
+import com.neovisionaries.i18n.CountryCode
+
+abstract class Product() {
+
+    @JsonIgnore
+    abstract fun getProductName(): String
+    @JsonIgnore
+    abstract fun getMarket(): CountryCode
+    @JsonIgnore
+    abstract fun getAddress(): Address?
```
bystam, line 13: You can also spell them like `abstract val address: Address?`, which is more in line with kotlin conventions.

src/main/kotlin/com/hedvig/underwriter/products/model/Product.kt
```diff
@@ -0,0 +1,14 @@
+package com.hedvig.underwriter.products.model
+
+import com.fasterxml.jackson.annotation.JsonIgnore
+import com.neovisionaries.i18n.CountryCode
+
+abstract class Product() {
```
bystam, line 6: If it has no implementations at all, could we make it simply an interface?

src/main/kotlin/com/hedvig/underwriter/products/service/Products.kt
```diff
@@ -0,0 +1,19 @@
...
+import kotlin.reflect.KClass
+
+class Products {
+
+    companion object {
+        fun getClass(productName: String): KClass<out Product> {
+
+            when (productName) {
+                "SwedishAccident" -> return SwedishAccident::class
+            }
```
bystam, line 14: Is there a reason we can't use jackson subtype info for this? I am thinking:

```kotlin

@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "productName")
@JsonSubTypes(
    JsonSubTypes.Type(value = SwedishAccident::class, name = "SwedishApartment")
)
interface Product {
  ...
}
```

bystam: Then we would get two-way json serialization simply from the declaration of the `Product` interface.

src/main/kotlin/com/hedvig/underwriter/model/QuoteData.kt
```diff
@@ -18,13 +19,72 @@ import java.util.UUID
...
+
+    @get:JsonIgnore
+    override val street: String?
+        get() = product.getAddress()?.street
+    @get:JsonIgnore
+    override val zipCode: String?
+        get() = product.getAddress()?.zipCode
+    @get:JsonIgnore
+    override val city: String?
+        get() = product.getAddress()?.city
```
bystam, line 73: Feels a bit weird to have `Product` have a `getAddress()` requirement, because address can be so ambiguous (address of home? signer? something else?). But I assume it's needed to fulfil this existing requirement on QuoteData?

src/main/kotlin/com/hedvig/underwriter/web/dtos/QuoteRequestDto.kt
```diff
@@ -43,3 +39,23 @@ data class QuoteRequestDto(
...
+            "norwegianHomeContents" -> return objectMapper.treeToValue(data, QuoteRequestData.NorwegianHomeContents::class.java)
+            "norwegianTravel" -> return objectMapper.treeToValue(data, QuoteRequestData.NorwegianTravel::class.java)
+            "danishHomeContents" -> return objectMapper.treeToValue(data, QuoteRequestData.DanishHomeContents::class.java)
+            "danishAccident" -> return objectMapper.treeToValue(data, QuoteRequestData.DanishAccident::class.java)
+            "danishTravel" -> return objectMapper.treeToValue(data, QuoteRequestData.DanishTravel::class.java)
+        }
+
+        val genericData = objectMapper.treeToValue(data, Products.getClass(type).java)
+        return QuoteRequestData.Generic(type, genericData)
+    }
```
bystam, line 60: Instead of a `JsonDeserializer`, shouldn't the existing declaration you removed also have worked, by simply moving it to the `QuoteRequestData` type and adding a default case? That combined with using the subtype annotations for `Product` as well should get rid of the need for manual type mapping like this.

src/main/resources/db/changelog/migrations/20210601-1600-generic-quotes.sql
```diff
@@ -0,0 +1,172 @@
...
+     address_zipcode varchar(20),
+     address_city varchar(255),
+     productType varchar(100) NOT NULL,
+     product jsonb NOT NULL
+);
+
+ALTER TABLE quote_revisions
+    ADD COLUMN quote_generic_data_id integer NULL REFERENCES quote_revision_generic_data(internal_id),
+    DROP CONSTRAINT single_quote_type,
+    ADD CONSTRAINT single_quote_type CHECK (
```
bystam, line 25: Holy crap, do we do this every time we add a quote type? 😅

bystam (commented):
Looks awesome! I added a few thoughts to it

PR #291, opened 2021-09-06
MX-261: Create a flow for signing quotes without having a member upfront

bystam:
# Jira Issue: [MX-261] 

## What?
- Expands the "Onboarding session" concept to encompass signing of insurance as well
- This flow aims to make Underwriter the somewhat designated onboarding service, trying to move dependency arrows in a single direction (underwriter and outwards) rather than having quite deep and cyclical dependencies
- The goal of this signing method is to have a single service where it's simple to answer the question: "What happens when the user signs?"
- It still leverages member-service to perform the BankID dance, but that dance is 100% stateless and has no side effects in member-service

## Why?
- Signing is a rather complex but also quite confusing flow in our system. It involves lots of services, mutating critical state in tandem, and calls going back and forth between services sometimes even cyclically
- Signing insurance is our main driver of conversion and revenue, so we should put time into making sure it's somewhat simple to understand and work with it
- We are also on a quest to make a `member` something that only actually appears once you actually... are a member from a business perspective (i.e. you have an insurance with us)

## Depends on
https://github.com/HedvigInsurance/member-registry-service/pull/538
https://github.com/HedvigInsurance/member-registry-service/pull/539

## API flow breakdown
The new signing flow can be described like this (see picture downstairs)

New concepts for the `OnboardingSession`:
- `signStatus` an enum describing the current status of the sign session
- `createdMemberId` to write down which member was created from this signing process

(This assumes there's an active onboarding session)
- Start signing
  - takes onboardingSessionId
  - returns either
    - "time to do BankID swap", sign status is `PENDING`
    - or "signed! This was simple sign", sign status is `READY`
- If SE BankID, member service has registered a quartz poller
  - This poller has the personal number and a "callback URL" injected
  - upon completion, it fires the results on the callback URL
  - If success, it sets everything up and marks the session with status SIGNED, if failure it is marked with FAILED
- Once the sign session is `SIGNED`, a member exists with contracts and all that
- Clients can poll the session state until it reaches either SIGNED or FAILED
- Once it is `SIGNED`, it can be consumed to create an `accessToken` (auth token)
  - This can only be done once for security purposes
  - This accessToken can be used to proceed the flow and do things like connecting payments
  - Consuming the sign session marks it as `COMPLETED`
- After consuming the session for an access token, the flow is complete

## New call structure
image

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [ ] Tested locally

[MX-261]: https://hedvig.atlassian.net/browse/MX-261


src/main/kotlin/com/hedvig/underwriter/service/MemberCreationHighway.kt
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * When running this code path, quotes are not bound to any particular "member" yet, but once
+ * they are signed, we create all the state necessary for the user to become a fully fledged member
+ * of Hedvig.
+ *
+ * This is a somewhat scary distributed "transaction" that mostly needs to happen atomically. There
+ * is currently no way for us to ensure that other than keeping this code path fail free.
+ */
+@Service
+class MemberCreationHighway(
```
bystam, line 41: This is not done yet FYI :)

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/MemberServiceClient.kt
```diff
@@ -30,12 +38,46 @@ import org.springframework.web.bind.annotation.RequestHeader
...
+    fun updateMember(
+        @PathVariable memberId: String,
+        @RequestBody body: UpdateMemberRequest
+    )
+
+    @PostMapping("/_/member/{memberId}/signed")
+    fun markMemberAsSigned(
+        @PathVariable memberId: String,
+        @RequestBody body: MemberSignedRequest
+    )
```
bystam, line 61: New endpoint

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/MemberServiceClient.kt
```diff
@@ -66,13 +108,10 @@ interface MemberServiceClient {
...
+    // De-coupled signing API
 
-    @DeleteMapping("/_/member/{memberId}")
-    fun deleteMember(
-        @PathVariable("memberId") memberId: String
-    ): ResponseEntity<*>
+    @PostMapping("/_/se-bankid/sign")
+    fun startSwedishBankIdSignSession(
+        @RequestBody body: SeBankIdSigningRequest
+    ): SeBankIdSigningResponse
```
bystam, line 116: New endpoint

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/MemberServiceClient.kt
```diff
@@ -30,12 +38,46 @@ import org.springframework.web.bind.annotation.RequestHeader
 @ConditionalOnProperty("hedvig.integrations.fakes", havingValue = "false", matchIfMissing = true)
 interface MemberServiceClient {
 
+    // member lifecycle
+
     @PostMapping("v2/member/helloHedvig")
-    fun createMember(): ResponseEntity<HelloHedvigResponseDto>
+    fun helloHedvig(): ResponseEntity<HelloHedvigResponseDto>
```
bystam, line 44: I renamed this `helloHedvig`, because it actually has more side effects than simply creating a member, and instead let `createMember` be the side-effect-less CRUD operation.

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
```diff
@@ -0,0 +1,187 @@
...
+            )
+        )
+        memberService.markMemberAsSigned(
+            memberId = memberId,
+            body = if (mandate != null)
+                MemberSignedRequest.SwedishBankId(mandate.referenceToken, mandate.signature, mandate.oscpResponse)
+            else
+                MemberSignedRequest.SimpleSign(data.personalNumber, data.countryCode)
+        )
+        notificationService.associateMemberWithOnboarding(
```
bystam, line 96: Not 100% sure the API will look like this. Customer.io is an ongoing thing

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
```diff
@@ -0,0 +1,192 @@
...
+        maybeRedeemCampaigns(memberId, quotes)
+        setUpInBotService(memberId, memberData)
+
+        return Result(memberId)
+    }
+
+    private fun setUpMember(data: MemberData, onboardingSession: OnboardingSession, mandate: SeBankIdMandate?): String {
+        val memberId = memberService.createMember(
+            body = CreateMemberRequest(acceptLanguage = onboardingSession.locale)
+        ).memberId.toString()
```
bystam, line 75: we accidentally made the `createMember` endpoint return the ID as an integer 😢

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
```diff
@@ -0,0 +1,192 @@
...
+ * This is a somewhat scary distributed "transaction" that mostly needs to happen atomically. There
+ * is currently no way for us to ensure that other than keeping this code path fail free.
+ */
+@Service
+class MemberSignedHighway(
+    private val quoteRepository: QuoteRepository,
+    private val memberService: MemberServiceClient,
+    private val productPricing: ProductPricingClient,
+    private val botService: BotServiceClient,
+    private val notificationService: NotificationServiceClient
```
bystam, line 50: Not sure what you feel about this. I sometimes feel like wrapping the external service in an Impl that implements a service interface is a bit bloated. It requires clicking 3 times to find the implementation, when you basically always just want to find what URL something is calling.

At the very least I think it would be nice to get rid of the outermost interface, or place it in the same file as the impl. Sometimes I feel like I have like 10 tabs open in IntelliJ where there's ACTUALLY only like 3 of them that contains code worth reading, the rest just being microscopic and inconsequential files to establish the pattern.

But mostly I would personally love just using the FeignClient directly. I feel like wrapping the client in a service does some harm in obfuscating what's actually going on underneath. I assume the goal is to kind of place an abstraction where the reader "shouldn't need to know whether it's a repository or an external service" underneath, but I quite often find that to be important information. It creates quite a lot of depth in the call stack where it could just be a lot flatter and more obvious.

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/dtos/MemberSignedRequest.kt
```diff
@@ -0,0 +1,23 @@
...
+
+import com.fasterxml.jackson.annotation.JsonSubTypes
+import com.fasterxml.jackson.annotation.JsonTypeInfo
+import com.neovisionaries.i18n.CountryCode
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonSubTypes(
+    JsonSubTypes.Type(MemberSignedRequest.SwedishBankId::class, name = "swedish_bank_id"),
+    JsonSubTypes.Type(MemberSignedRequest.SimpleSign::class, name = "simple_sign")
+)
```
bystam, line 11: I just realised now that I'm super used to snake-casing heterogenous type declarations like this, but we don't do that elsewhere in underwriter.

Should I change these to "SwedishBankId" and "SimpleSign" maybe?

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/dtos/Address.kt
```diff
@@ -2,10 +2,10 @@ package com.hedvig.underwriter.serviceIntegration.memberService.dtos
...
 
-class Address(
+data class Address(
     @Masked val street: String,
     val city: String,
     val zipCode: String,
-    val apartmentNo: String,
-    val floor: Int
+    val apartmentNo: String? = null,
+    val floor: Int? = null
```
bystam, line 10: These are all nullable in their destination service (member-service), so this should be safe

src/main/kotlin/com/hedvig/underwriter/web/QuoteSignController.kt
```diff
@@ -0,0 +1,118 @@
...
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.server.ResponseStatusException
+
+@RestController
+@RequestMapping(
+    "/_/quote-signing"
```
bystam, line 23: TODO this should be a public endpoint for giraffe?

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
bystam: I realise now that I think I need to run this earlier, because the other services called from here also depend on notification-service and will need this state.

bystam: Yeah so this is a TODO hehe

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
```diff
@@ -0,0 +1,192 @@
...
+ *
+ * When running this code path, quotes are not bound to any particular "member" yet, but once
+ * they are signed, we create all the state necessary for the user to become a fully fledged member
+ * of Hedvig.
+ *
+ * This is a somewhat scary distributed "transaction" that mostly needs to happen atomically. There
+ * is currently no way for us to ensure that other than keeping this code path fail free.
+ */
+@Service
+class MemberSignedHighway(
```
mightyost, line 45: I see that this class is locked to Sweden with regards to `SeBankIdMandate` and `MemberSignedRequest.SwedishBankId`, is it possible to make this more generic to fit other markets in the future?

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
```diff
@@ -0,0 +1,192 @@
...
+        val birthDate: LocalDate,
+        val email: String?,
+        val phoneNumber: String?,
+        val address: Address?
+    ) {
+        companion object {
+            // The member data is modeled inside the list of quotes.
+            // In order to turn them into member information we need to make
+            // some pretty paranoid extraction.
+            fun extract(quotes: List<Quote>) = MemberData(
```
mightyost, line 164: Wonder if we should have a validation check close to endpoint where all these properties are enforced to be the same, or updated to be the same?

mightyost: I see that we have this logic in more places to extract a property from a list of quotes, maybe have it reusable in an extension function...

src/main/kotlin/com/hedvig/underwriter/service/QuoteSigningService.kt
```diff
@@ -0,0 +1,112 @@
...
+                    status = SignSessionStatus.PENDING,
+                    bankIdAutostartToken = response.autostartToken
+                )
+            }
+            Market.NORWAY,
+            Market.DENMARK -> {
+                val onboardingSession = onboardingSessionRepository.fetch(onboardingSessionId)!!
+                val result = memberSignedHighway.runMemberCreation(
+                    quotes, onboardingSession, null
+                )
```
mightyost, line 64: Question. Would it be possible if `startSigning` just initiated the signing and not in the case of NO/DK (simple sign?) also is creating member, contracts and redeem. I.e. that this logic is consolidated in `completeSigning` for all types of signing flows instead?

That means we will treat all signing types as async where the flow is:
- start signing
- complete signing

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/dtos/MemberSignedRequest.kt
mightyost: Is this not dictated by the member service and its protocol? I.e. if snake-case is used there then UW has to adapt...

But I also like snake case better in general for this.

mightyost (commented):
Nice work, I like!!

One thing that I wonder if possible and that would feel more future prove and elegant when adding new markets and/or new signing integrations. Is it possible to have a totally generic asynchronous signing api in MemberService which clients can call to sign whatever without having to know which integration to use?

Simplistic example:
```
initiateSiging(nationalIdentification, market, text, resultUrl): sessionId

getToken(sessionId): token
```
The client calls `initiateSiging` when to sign anything and gets back a sessionId. The `resultUrl` is triggered by MemberService when user has signed with sessionId and result successful or failed. The client calls `getToken` to get token.

If it is possible to make it totally generic I think UW will look nicer and not have to keep track of all specific signing stuff...

This would also be very reusable by other services/flows needing to sign stuff....

src/main/kotlin/com/hedvig/underwriter/model/SignSessionStatus.kt
```diff
@@ -0,0 +1,28 @@
...
+enum class SignSessionStatus {
+    /**
+     * This session is started, but still awaiting some sign mechanism to happen (like BankID)
+     */
+    PENDING,
+
+    /**
+     * The signing has succeeded, and is ready to be consumed for an access token.
+     */
+    READY,
```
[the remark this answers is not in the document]
bystam: Oof... I think I might like `SIGNED`? The reason I "dislike" the ones about token readiness is that the status also implies so much else has also happened.

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
bystam: Since none of that is persisted, I find such a thing to be a bit of premature optimization in the data modeling. Since we have no idea what that will look like, I personally find it better to just let it be dumb for now so that there's as much room as possible for changes with the new markets and signing flows.

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
bystam: If it's just a transformation of HTTP errors, that can be done using feign configs too. That's the library-supported way of doing it. But yeah, if there's more complex transformations than that, sure.

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
```diff
@@ -0,0 +1,192 @@
...
+    ): Result {
+        val memberData = MemberData.extract(quotes)
+        if (memberService.checkIsSsnAlreadySignedMemberEntity(memberData.personalNumber).ssnAlreadySignedMember) {
+            throw RuntimeException("Member personal number already signed, should not have ended up here.")
+        }
+
+        val memberId = setUpMember(memberData, onboardingSession, mandate)
+
+        createContractsAndAgreements(memberId, quotes, mandate)
+        maybeRedeemCampaigns(memberId, quotes)
```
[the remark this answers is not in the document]
bystam: I guess it's a matter of perspective. In my mind, the way our system works, having a member without a contract (from now on) is a little bit of an illegal state. So they are very connected in that sense.

The fact that there technically are endpoints to create a member without setting it up is an implementation detail to me.

But you're right, I could probably name things better here. I still kind of prefer having it as a single method though. I want to give the impression that this is an atomic operation of sorts.

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
```diff
@@ -0,0 +1,192 @@
...
+        setUpInBotService(memberId, memberData)
+
+        return Result(memberId)
+    }
+
+    private fun setUpMember(data: MemberData, onboardingSession: OnboardingSession, mandate: SeBankIdMandate?): String {
```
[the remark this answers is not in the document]
bystam: The variable or class? The reason the class isn't called `Member` is because when it's instantiated in the code, there is no member. It is a projection of the state inside the Quotes that **will** belong to a member.

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
```diff
@@ -0,0 +1,192 @@
...
+                hasSigned = true
+            )
+        )
+    }
+
+    private data class MemberData(
+        val firstName: String,
+        val lastName: String,
+        val personalNumber: String,
+        val countryCode: CountryCode,
```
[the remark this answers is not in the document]
bystam: I like :)

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
bystam: > I see that we have this logic in more places to extract a property from a list of quotes, maybe have it reusable in an extension function...

I mean... maybe. I personally think we shouldn't model data inside a "list of quotes" this way. It feels like there is some kind of parent data type missing that should contain it instead.

But more specifically, to me this `MemberData` class basically **is** the extracted, reusable tool. If we want we could move it to some new file and called it `QuoteMemberProperties` or something.

I am not a big fan of extensions on `List`s that we've done so far... they are very confusing to read in my mind.

> Wonder if we should have a validation check close to endpoint where all these properties are enforced to be the same, or updated to be the same?

That's outside of scope for this PR I think :P

src/main/kotlin/com/hedvig/underwriter/service/QuoteSigningService.kt
bystam: At first I thought that you meant having a separate endpoint, but I think you mean that we could just as easily just call the `completeSigning` method from the endpoint immediately in the case of simple sign?

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/dtos/MemberSignedRequest.kt
bystam: Right, good point. This is an outgoing request. I'll check MS.

bystam:
> One thing that I wonder if possible and that would feel more future prove and elegant when adding new markets and/or new signing integrations. Is it possible to have a totally generic asynchronous signing api in MemberService which clients can call to sign whatever without having to know which integration to use?

So, in fact, this is kind of what it looks like right now - and I purposefully decided to move **away** from that. I will try to explain why.

Firstly, the whole art of "signing" is something that I feel is underwriters job, not member-service. "Signing" is not a very well defined thing from a technical standpoint. It's just something that happens at the end of the onboarding that has a bunch of side effects in our system. The new endpoint I built in member-service for signing with BankID literally has zero side effects inside member service. I just placed it in there because member-service already had Quartz set up. If I were bold I would've moved all of that here and removed the concept of signing from member-service altogether. I think that's a nicer future than what we have now. Because I feel that the fact that member-service is involved at all is purely ritualistic.

The "generic signing API that doesn't care about the sign method" to me already exists, that's what it's in this PR and it's in underwriter. Delegating that part to member-service just feels like hopping between services "just because".

I think it's very hard to bet on what signing will look like in future markets. I am pretty sure betting on some generic API in MS or somewhere else will quite likely be premature API design that has to be rebuilt when we know how it works in France.

> 
> Simplistic example:
> 
> ```
> initiateSiging(nationalIdentification, market, text, resultUrl): sessionId
> 
> getToken(sessionId): token
> ```
> 
> The client calls `initiateSiging` when to sign anything and gets back a sessionId. The `resultUrl` is triggered by MemberService when user has signed with sessionId and result successful or failed. The client calls `getToken` to get token.

I am not sure it will make sense to bet on this structure with new markets necessarily. I bet the `nationalIdentification` concept will go away with future markets, and only actually exist for SE/DK/NO anyway. So going on a deep dive to hard-wire that into the API will probably just mean having to remodel it once we go to France.

> If it is possible to make it totally generic I think UW will look nicer and not have to keep track of all specific signing stuff...

But then all of those things would have to happen in member-service instead. And then there's just more service jumps, more DTOs and all of that. I think the end result would give the appearance that there's more that's happening than is actually true. "Simple sign" is just slang for "no signing at all really", and if we turn that into 

> This would also be very reusable by other services/flows needing to sign stuff....

Right now I think betting on that would mean building something that won't be used. Signing with BankID is an actual signing process, but simple sign is not. If we ever want to have a signing process inside the apps or something for something else (like maybe submit claim?), then the odds of that using the same SE BankID/SimpleSign split across the markets to me feels super unlikely. SimpleSign is not really a Norway/Denmark standard for signing, it's a conversion hack because signing with real NO/DK bankid was so ugly.

-----

I think the existing features we have make way more sense if you think of "Simple Sign" as having the name "No Sign". Because there is nothing actually signed. No signatures, no proof of identification, nothing. It's just a press of a button that says "signed". So to me, if we were to design an elaborate umbrella-concept for "signing", chances are we're trying to mix oil and water a bit.

bystam (commented):
Thanks a bunch for the review! I reacted to almost all of the comments and the big headline questions as well :)

src/main/kotlin/com/hedvig/underwriter/model/SignSessionRepository.kt
```diff
@@ -1,8 +1,39 @@
 package com.hedvig.underwriter.model
 
 import java.util.UUID
+import org.jdbi.v3.core.Jdbi
+import org.jdbi.v3.sqlobject.kotlin.attach
+import org.springframework.stereotype.Component
 
-interface SignSessionRepository {
```
tobiasbexelius, line 5: 🎉

src/main/kotlin/com/hedvig/underwriter/model/SignSessionStatus.kt
```diff
@@ -0,0 +1,28 @@
...
+
+    /**
+     * The signing has failed.
+     */
+    FAILED,
+
+    /**
+     * The default case when not running the new signing flow.
+     */
+    NONE
```
tobiasbexelius, line 27: Is this needed?

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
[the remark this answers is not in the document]
tobiasbexelius: Agree that naming could be improved, but I do think that it should be a single operation. 

Is highway a conventional naming pattern? I also like `createAndSetupMember`

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
tobiasbexelius: Maybe feign is flexible enough that we can declare it as a string instead in this service without changing the other end?

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
```diff
@@ -0,0 +1,192 @@
...
+                data.lastName,
+                data.personalNumber,
+                data.countryCode,
+                data.address,
+                data.email,
+                data.phoneNumber,
+                data.birthDate
+            )
+        )
+        memberService.markMemberAsSigned(
```
tobiasbexelius, line 89: We should merge these three calls at some point

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
```diff
@@ -0,0 +1,192 @@
...
+        return memberId
+    }
+
+    private fun createContractsAndAgreements(memberId: String, quotes: List<Quote>, mandate: SeBankIdMandate?) {
+        val request = CreateContractsRequest.from(
+            quotes = quotes,
+            memberId = memberId,
+            signedRequest = mandate?.let { SignRequest(it.referenceToken, it.signature, it.oscpResponse) }
+        )
+        val creations = productPricing.createContract(request, token = null)
```
tobiasbexelius, line 109: I was a bit surprised to see the token parameter here. Do you know what that is for?

src/main/kotlin/com/hedvig/underwriter/service/QuoteSigningService.kt
```diff
@@ -0,0 +1,112 @@
...
+    private val quoteService: QuoteService,
+    private val onboardingSessionRepository: OnboardingSessionRepository,
+    private val memberSignedHighway: MemberSignedHighway,
+    private val apiGatewayService: ApiGatewayService
+) {
+
+    fun startSigning(quoteIds: List<UUID>, onboardingSessionId: UUID, ipAddress: String): StartSigningResponse {
+        val quotes = quoteService.getQuotes(quoteIds)
+        // Quotes are assumed to have a uniform market
+        val market = quotes.map { it.market }.toSet().single()
```
tobiasbexelius, line 31: Nice way to enforce that constraint!

src/main/kotlin/com/hedvig/underwriter/service/QuoteSigningService.kt
```diff
@@ -0,0 +1,112 @@
...
+        // Quotes are assumed to have a single personal number
+        val personalNumber = quotes.map { it.ssn }.toSet().single()
+
+        if (memberService.checkIsSsnAlreadySignedMemberEntity(personalNumber).ssnAlreadySignedMember) {
```
tobiasbexelius, line 35: What's up with these names?

src/main/kotlin/com/hedvig/underwriter/service/QuoteSigningService.kt
```diff
@@ -0,0 +1,112 @@
...
+            throw ResponseStatusException(HttpStatus.FORBIDDEN)
+        }
+
+        when (market) {
+            Market.SWEDEN -> {
+                val session = repository.insert(
+                    SignSession(
+                        UUID.randomUUID(), onboardingSessionId, SignSessionStatus.PENDING, quoteIds = quoteIds
```
tobiasbexelius, line 43: Cosmetic: parameter names and separate lines makes it prettier IMO

src/main/kotlin/com/hedvig/underwriter/service/QuoteSigningService.kt
```diff
@@ -0,0 +1,112 @@
...
+                    )
+                )
+                val response = memberService.startSwedishBankIdSignSession(
+                    SeBankIdSigningRequest(
```
tobiasbexelius, line 47: Same

src/main/kotlin/com/hedvig/underwriter/service/QuoteSigningService.kt
tobiasbexelius: I like the idea of keeping the flows more uniform, i.e. calling completeSigning from here, if that makes sense

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/MemberServiceClient.kt
tobiasbexelius: Do we need the CRUD operations?

tobiasbexelius (commented):
Very cool!

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
bystam: Looks like it's a value sent when a quote is signed from Hope. `token` is a bit ambiguous, but the value sent to it is the property written on `Quote` called `signFromHopeTriggeredBy`

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
bystam: Yes I had the same thought the other day. There should be a way to simply create a member that's all done.

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
bystam: Interesting thought... Don't think that would be Feign though but rather Jackson. I could do a quick test to see if that's the case.

bystam: image
image

Turns out it should work :D I'll change it

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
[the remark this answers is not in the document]
bystam: I remember people using "highway" as a fun way to describe a "sequence of stuff that's happening back-to-back".

I am 100% up for changing it, but I would prefer not to call is "service", as I think "service" is so non-descriptive. I think it's so hard to understand code when services wrap other services. I don't see this as a service (as in something that caters to a domain), but rather just a way to give an important function some kind of descriptive name.

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/MemberServiceClient.kt
bystam: I mean, it would be way better if we had a way to create a member that had all of the important state upfront.

That would be nicer to use in rapio too, right? I will get to it immediately honestly.

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
tobiasbexelius: 🎉

src/main/kotlin/com/hedvig/underwriter/model/SignSessionStatus.kt
bystam: It was either that or making it nullable, since this is applied to an existing table that needed a default value for all the rows that are already in there.

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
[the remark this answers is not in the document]
mightyost: The variable name I ment. The name "data" gives not much context to me, "member" does ...

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
mightyost: 👍

mightyost (approved):

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
bystam: https://github.com/HedvigInsurance/member-registry-service/pull/539 :)

src/main/kotlin/com/hedvig/underwriter/service/QuoteSigningService.kt
bystam: The names of the methods? Well, it checks if the SSN values is already in some `SignedMemberEntity`.

mightyost:
> > One thing that I wonder if possible and that would feel more future prove and elegant when adding new markets and/or new signing integrations. Is it possible to have a totally generic asynchronous signing api in MemberService which clients can call to sign whatever without having to know which integration to use?
> 
> Firstly, the whole art of "signing" is something that I feel is underwriters job, not member-service. "Signing" is not a very well defined thing from a technical standpoint. It's just something that happens at the end of the onboarding that has a bunch of side effects in our system. The new endpoint I built in member-service for signing with BankID literally has zero side effects inside member service. I just placed it in there because member-service already had Quartz set up. If I were bold I would've moved all of that here and removed the concept of signing from member-service altogether. I think that's a nicer future than what we have now. Because I feel that the fact that member-service is involved at all is purely ritualistic.
> 
> The "generic signing API that doesn't care about the sign method" to me already exists, that's what it's in this PR and it's in underwriter. Delegating that part to member-service just feels like hopping between services "just because".
> 
> I think it's very hard to bet on what signing will look like in future markets. I am pretty sure betting on some generic API in MS or somewhere else will quite likely be premature API design that has to be rebuilt when we know how it works in France.

Hmm, I guess we look at this a little bit differently :). Yes, agree that it is Underwriter's job facilitate and sign quotes. But to me the actual external signing integrations could preferable be managed in its own dedicated micro service with a nice generic api on top for Underwriter (or other services) to use. Does not feel so far fetched that we want to sign other things in other parts of the systems in the future. 

Lets revisit this when we have added more integrations :).

bystam:
> Hmm, I guess we look at this a little bit differently :). Yes, agree that it is Underwriter's job facilitate and sign quotes. But to me the actual external signing integrations could preferable be managed in its own dedicated micro service with a nice generic api on top for Underwriter (or other services) to use. Does not feel so far fetched that we want to sign other things in other parts of the systems in the future.

I think we do look at it differently. In my mind, there is no well-defined and broadly understood concept called "signing". Signing is more of a point-in-time in the UI than it has a technical definition. That's how I see it. Therefore I am not very eager to create a bunch of abstractions betting that we have some company-wide understanding of what signing means.

Regardless, if we do end up having some kind of rigorous understanding of it, I don't see why `member-service` would host it necessarily. Right now, "signing quotes" is part of the "purchase journey". And from my perspective, `underwriter` might as well be called the `purchase-journey-service`, because that's what it does mostly ^^

> Lets revisit this when we have added more integrations :).

Yes! At least none of this is Public API, so we can easily change it later if we want to.

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/MemberServiceClient.kt
bystam: https://github.com/HedvigInsurance/member-registry-service/pull/539

src/main/kotlin/com/hedvig/underwriter/web/OnboardingSessionController.kt
```diff
@@ -45,4 +83,56 @@ class OnboardingSessionController(
...
+        val tokenCreationUrl: String?
+    )
+
+    data class TokenCreationResponse(
+        val accessToken: String
+    )
+}
+
+@RestController
+class OnboardingSeBankIdCallbackController(
```
bystam, line 100: Place this endpoint on a different controller, because it should not be public :)

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
```diff
@@ -0,0 +1,120 @@
...
+            throw ResponseStatusException(HttpStatus.FORBIDDEN)
+        }
+
+        val quotes = quoteRepository.findByOnboardingSessionId(onboardingSessionId)
+        if (quotes.isEmpty()) throw ResponseStatusException(HttpStatus.FORBIDDEN)
+
+        // Quotes are assumed to have a uniform market
+        val market = quotes.map { it.market }.toSet().single()
+        // Quotes are assumed to have a single personal number
+        val personalNumber = quotes.map { it.ssn }.toSet().single()
```
mightyost, line 47: Is this dangerous?

I guess `ssn` is mandatory when creating quotes in Sweden but not in other markets like Norway...

```
setOf(null, "12345").toSet().single()
```
...will throw exception, maybe filter out nulls first?

tobiasbexelius (approved):
Cool! Let's add @deprecate to the sign session things so that we remember to remove it later

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
```diff
@@ -0,0 +1,200 @@
...
+                        Address(
+                            it.street!!,
+                            it.city!!,
+                            it.zipCode!!,
+                            (it as? DanishHomeContentAddressData)?.apartment,
+                            (it as? DanishHomeContentAddressData)?.floor?.toIntOrNull()
+                        )
+                    }
+                    .toSet()
+                    .singleOrNull()
```
mightyost, line 192: Doing `quotes.map { it.xxx }.toSet().single()`on lot of nullable fields feels dangerous. Maybe filter out the nulls before converting to set to be on the safe side?

I know that you do not like to add extensions on lists but since I see this (bloating) pattern in many places, maybe add a generic function to extract a single non-null field if any, e.g. something like `quotes.singleValue(::ssn)`?

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/MemberServiceClient.kt
```diff
@@ -66,13 +95,17 @@ interface MemberServiceClient {
...
+    fun startSwedishBankIdSignSession(
+        @RequestBody body: SeBankIdSigningRequest
+    ): SeBankIdSigningResponse
+
+    // User lifecycle
+
+    @PostMapping("/_/user")
+    fun createUser(
+        @RequestBody body: CreateUserRequest
+    ): CreateUserResponse
```
mightyost, line 110: Just a curious question. I see member and person in the api. What is an user?

mightyost (approved):
Looks good. I had some comments around `.single()` that feels a bit uncomfortable...

mightyost:
Also maybe needed?

Make sure that the quotes to sign are validated in the new `BundleQuotesService.validateBundling(memberId: String?, quotes: List<Quote>)` method...

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/MemberServiceClient.kt
mightyost: I guess its the onboarding user that is neither a member or person yet, right?

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
bystam: Wait what - how is SSN not mandatory? What do we use to create the account later then?

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/MemberServiceClient.kt
bystam: The user is the login - our incremental attempt at separating the human from the account.

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
mightyost: Sorry, I was thinking too fast. We always have a ssn when signing. But not when creating quotes.

src/main/kotlin/com/hedvig/underwriter/service/MemberSignedHighway.kt
bystam: > Doing quotes.map { it.xxx }.toSet().single()on lot of nullable fields feels dangerous. Maybe filter out the nulls before converting to set to be on the safe side?

I was under the impression that they must not be null at this point. I basically took this code from some other place where the same code was run but only on the **first** of the quotes. Then I decided to improve upon it by going from first to checking all of them. These actually use non-nullable getters on Quote too:

image

So I don't think the code does what it looks like it does on Github.

------

I am fine with extensions on lists as long as they operate on list elements generically. I just dislike doing things like:

```kotlin
fun List<MyVerySpecificObject>.methodThatMakesItLookLikeSomethingOtherThanAList() { ... }
```

PR #297, opened 2021-09-09
MX-269: create an 'onboarding' data type that can be used to keep track of onboarding-only state

bystam:
# Jira Issue: [MX-269] 

## What?
- add a new data type called `OnboardingSession`
- It will be created at the start of the onboarding, populated with the correct language/market
- Its ID can be used to identify non-signed onboarding "leads" in things like customer.io for instance

## Why?
- We need ways to identify an "onboarding session" without a memberId
- The quoteIds are no good, because they are multiple and don't have a strong connection

## What's going on here?
The MX team is working to de-couple the "member" concept away from pre-onboarded users. In order words, we're trying to make a "member" something that appears when signing. This in order to have way better control over the member lifecycle.

As part of this journey, we have considered `underwriter` to be closest thing we have to an "onboarding service". We also consider creating "yet another service" is not very helpful at this moment. We would rather work to improve the relationship between the services we already have. Therefore, we're trying to ever so slightly consolidate the "onboarding" flow in underwriter, where the biggest piece of onboarding thing already happens (quoting).

More information here: https://www.notion.so/hedviginsurance/Decouple-member-from-onboarding-e6fe6120275049f9905389080ff51d1b

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [ ] Tested locally

[MX-269]: https://hedvig.atlassian.net/browse/MX-269


src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.http.HttpStatus
+import org.springframework.web.bind.annotation.GetMapping
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.PostMapping
+import org.springframework.web.bind.annotation.RequestBody
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.server.ResponseStatusException
+
+@RequestMapping("/onboarding")
```
bystam, line 15: Not underscored because I intend for this type to be publicly accessible through giraffe.

bystam: Maybe it should be `/underwriter/onboarding` if we ever want to move towards uniform path prefixes for a service?

vonElfvin: What happened to external? 🤔 I honestly don't know how one should do this
```suggestion
@RequestMapping("/external/onboarding")
```

vonElfvin (commented):
Very neat 👍 

Some thoughts that came to my head:
1. How will we be able to connect this one to a member *after* sign?
2. How will this relate to a `User` if we were to setup authentication in the onboarding later?
3. How will this relate to connecting payments if we were to set that up in onboarding later?

I think 2. and 3. are impossible to answer directly today, but the first one might 🤷

src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
bystam: I don't recall seeing that pattern anywhere? From what I can see in the prod-env configmap, there is no discernible pattern at all between public endpoints. The only thing they have in common is to not have the underscore prefix.

I am previously used to letting each service have a single path prefix that denotes its domain, and only have a single redirect entry for that entire path. So like `path: /underwriter/** -> url: underwriter/`

bystam:
> 1. How will we be able to connect this one to a member _after_ sign?

You're right, we should probably have a nullable "memberId" column that we can populate upon sign. Not because we'll need it for business logic, but it could be helpful for traceability purposes.

> 2. How will this relate to a `User` if we were to setup authentication in the onboarding later?

Impossible to say at this stage. That depends on what we do then I think.

> 3. How will this relate to connecting payments if we were to set that up in onboarding later?

If we change things so that you're forced to connect payments BEFORE signing/creating member then I think this type is probably a decent place to put it. You could have a `payment_connection_id` column on it that you fill during onboarding, and then when everything is done you flush that ID to a properly set up entity in payment-service (or wherever that stuff will reside)

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.neovisionaries.i18n.CountryCode
+import java.util.UUID
+import org.jdbi.v3.core.Jdbi
+import org.jdbi.v3.sqlobject.customizer.Bind
+import org.jdbi.v3.sqlobject.kotlin.attach
+import org.jdbi.v3.sqlobject.statement.SqlQuery
+import org.jdbi.v3.sqlobject.statement.SqlUpdate
+import org.springframework.stereotype.Component
+
+data class Onboarding(
```
tobiasbexelius, line 12: Are we not able to use `@Entity` with liquibase?

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
```diff
@@ -0,0 +1,49 @@
...
+    @SqlQuery("""
+         SELECT * FROM onboardings WHERE id = :id
+    """)
+    fun fetch(
+        @Bind id: UUID
+    ): Onboarding?
+}
+
+@Component
+class OnboardingRepository(
```
tobiasbexelius, line 39: And no `@Repository`?

src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
tobiasbexelius: I prefer it the way it is FWIW

src/main/resources/db/changelog/changelog-master.yml
```diff
@@ -104,3 +104,6 @@ databaseChangeLog:
   - include:
       file: ./migrations/20210824-1030-add-swedish-accident.sql
       relativeToChangelogFile: true
+  - include:
```
tobiasbexelius, line 107: It would be cool if we could have automatic migrations for adding new tables, but force manual migration for updating/dropping

tobiasbexelius (commented):
Great! But I would really prefer if we could use @Entity and @Repository 🤔

src/main/resources/db/changelog/changelog-master.yml
bystam: I think it's hard to mix the two, because how would you keep track of which thing would happen when? The code only contains the "current state", whereas the changelog has a sequence.

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
bystam: @tobiasbexelius underwriter does not have `spring-data`, so we don't have that dependency.

Adding it became a hot topic recently discussed in #platform.

https://hedviginsurance.slack.com/archives/C9K7XRQ0G/p1630308935002400
https://hedviginsurance.slack.com/archives/C9K7XRQ0G/p1630309062002700

tobiasbexelius: I see. That's too bad, but let's tackle that another time then

tobiasbexelius (approved):

src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
vonElfvin: I think it was you who wrote `external` @bystam in `product-pricing`?

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
```diff
@@ -0,0 +1,49 @@
...
+import org.jdbi.v3.sqlobject.customizer.Bind
+import org.jdbi.v3.sqlobject.kotlin.attach
+import org.jdbi.v3.sqlobject.statement.SqlQuery
+import org.jdbi.v3.sqlobject.statement.SqlUpdate
+import org.springframework.stereotype.Component
+
+data class Onboarding(
+    val id: UUID,
+    val locale: String,
+    val market: CountryCode
```
vonElfvin, line 15: Having "market" as a CountryCode feels a bit weird to me. I think we never ever store the word "market" in any database which I think is a good thing. I guess we could figure out the `CountryCode` given the locale instead?

src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    @PostMapping
+    fun create(@RequestBody body: CreateInput): CreateOutput {
+        val id = UUID.randomUUID()
+        repository.insert(id, body.locale, body.market)
+        return CreateOutput(id)
+    }
+
+    data class ApiOnboarding(
```
vonElfvin, line 36: Why the word "Api"?

src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
```diff
@@ -0,0 +1,50 @@
...
+        val id: UUID,
+        val locale: String,
+        val market: CountryCode
+    )
+
+    data class CreateInput(
```
vonElfvin, line 42: CreateOnboardingRequest?

src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
```diff
@@ -0,0 +1,50 @@
...
+        val locale: String,
+        val market: CountryCode
+    )
+
+    data class CreateOutput(
```
vonElfvin, line 47: CreateOnboardingResponse?

vonElfvin (commented):
Neat concept, just a bit conflicted on the word "market" 🤔

src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
vonElfvin: Also, how does an onboarding relate to a quote? I guess they could be completely decoupled but that could be hard when debugging/tracking in analyics

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
tobiasbexelius: I'm not a big fan of using locale to determine market. To me locale feels more like something that determines the language in the UI, not something to base business logic on

tobiasbexelius: I'm not sure what it's supposed to be used for though 😄

src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
bystam: I always felt Request/Response were a bit off as names for the body, because in my mind the entire HTTP call is the Request (including url, method, headers, body), whereas this is just the body.

The reason I don't keep the word `Onboarding` in there is because I find that's evident from the context. This is nested inside `OnboardingController`, so "obviously" it's the input to create an `Onboarding`, and then it's nice to get rid of visual clutter in my mind.

bystam: I also think `Request` and `Response` are such similar words. It's more easy on my eyes to differentiate between the words `Input` and `Output` because of the I vs O ^^

src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
bystam: > Why the word "Api"?

Otherwise I would have to refer to the `Onboarding` data model by its qualified name including package. And I just now realise that it could've just been called `Dto` of course 🤔

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
bystam: I agree with Tobias. Locale to me is a language identifier. It doesn't even have to contain a country. "en" is a legal locale, specifying "the english language, no particular flavour".

I kind of don't like having `enum class Market` everywhere, because I find it to be such a coupled design. I think many services would feel better if markets were not modeled as closed sets, because it would force us to not hard-wire too much of our subsystems (services) as knowing that "yes, there are exactly 3 markets".

Maybe this is an instance where it should be stricter. I just generally feel like having `enum class Market { SWEDEN ... }` feels like re-inventing country codes but without its nice, portable format. At least I think we should consider moving towards naming them according to a 2-letter ISO code imo.

src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
bystam: I tried searching for `external` in the request mappings of most our big services, and couldn't find anything 🤔 ? I also don't see that word named anywhere in `prod-env/prod/prod-cluster/default/api-gateway/configmap.yml` for instance.

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
bystam: Or rather, `Locale` is not a language identifier per se, but it's a very loose concept that does not safely identify either language or country. From the Javadoc it says:

> A Locale object represents a specific geographical, political, or cultural region.

So... I mean. You can use it for lots of things. But I think it's most neatly leveraged to determine language and how to format things like dates, numbers etc.

src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
bystam: > Also, how does an onboarding relate to a quote? I guess they could be completely decoupled but that could be hard when debugging/tracking in analyics

It's not impossible that we will add that, but for now we have left it out. If we do add it, it would probably be part of what Tobias is working on. This was added in its own PR because both he and I will need to access this entity in parallel on both our branches.

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
mightyost: In general I like using the more "standardised" `CountryCode` class and stop using custom enums in our services. With it you also get build in support for resolving locales, timezones, currency etc.

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
mightyost, line 12: `Onboarding` is a bit ambiguous to me. I think `OnbardingUser` (I guess it's an user) would be better.

src/main/kotlin/com/hedvig/underwriter/web/OnboardingController.kt
mightyost: I have no problem with `Request/Response` ;), I also thinks it is better to follow the general pattern in UW than to create a new, or?

src/main/resources/db/changelog/changelog-master.yml
mightyost: I like to be 100% Liquidbase. Total control. :).

src/main/resources/db/changelog/migrations/20210909-1430-add-onboardings.sql
```diff
@@ -0,0 +1,13 @@
+--liquibase formatted.sql
+
+--changeset bystam:20210909-1430-add-onboardings.sql
+
+CREATE TABLE onboardings
+(
+    id               UUID PRIMARY KEY,
+    locale           TEXT NOT NULL,
+    market           TEXT NOT NULL,
+    result_member_id TEXT DEFAULT NULL
```
mightyost, line 10: Maybe use a `varchar(2)` for market to get an extra safety belt. 

Isn's `result_member_id` an uuid?

`result_member_id` could just be `member_id`, would work for me but is more of a cosmetic thing.

And as commented above, I think this table should be `onboarding_users` instead...

mightyost (commented):
I cannot really see the connection to Underwriter with this alone. But I guess it will be incorporated in quotes etc later?

Who will be consumers of `Onboarding`?

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
bystam: I... am not... entirely sure. I have viewed it more as the onboarding itself. If the same human being exits the onboarding flow and starts again, it will become a new `Onboarding`. So to me it's not reeeallly a user in that sense. We will make not attempt to stick it to the same person.

bystam:
> I cannot really see the connection to Underwriter with this alone. But I guess it will be incorporated in quotes etc later?
> 
> Who will be consumers of `Onboarding`?

This is a foundation PR that will be used in two subsequent ones currently worked on by me and @tobiasbexelius.

It will be used to keep track of a bunch of data related to the onboarding that we can't stick into the quotes themselves, as well as serve as a way to correlate multiple quotes somehow (outside of memberId, there is no way to detect that 2 quotes actually go together when creation both in the same onboarding).

The immediate need for this comes from us having to use something to keep track of people in customer.io.

The reason it's placed in `underwriter` is, as I wrote in the PR description, that I personally view underwriter as being the closest thing we have to an "onboarding service". It's currently the thing that both Rapio and Web/App onboardings fire requests towards during onboarding, that then internally flushes a bunch of state to other system to set things up. Whether intentional or not, I think underwriter is our `onboarding-service`.

src/main/resources/db/changelog/migrations/20210909-1430-add-onboardings.sql
bystam: > Maybe use a `varchar(2)` for market to get an extra safety belt.

Yeah sounds nice

> Isn's `result_member_id` an uuid?

Nope, it will be the regular 9 digit member id

> `result_member_id` could just be `member_id`, would work for me but is more of a cosmetic thing.

I added `result` to highlight the fact that it's null by default and then after a while will be given a member_id if the onboarding concludes. It's not tagged by one from the start. In my head, simply spelling `member_id` kind of makes it look like it was **created by that member**, but it's rather the other way around.

> And as commented above, I think this table should be `onboarding_users` instead...

I answered that thing above. Let's see if you agree with my reasoning :)

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
mightyost: Aha, puts more water on my grinder then ;), oki so maybe `OnboardingSession` is more accurate then?

bystam: Hmm... but maybe I actually have the wrong idea here. Maybe we would place this ID in local storage of the browser and actually consider it the user itself.

bystam:
Ping @mightyost I answered your comments :)

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
bystam: `OnboardingSession` is pretty nice. I will change to that

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
vonElfvin: Can't we call it `country` or `countryCode` instead if so?

vonElfvin (approved):
LGTM 👍

src/main/kotlin/com/hedvig/underwriter/model/OnboardingModel.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.neovisionaries.i18n.CountryCode
+import java.util.UUID
+import org.jdbi.v3.core.Jdbi
+import org.jdbi.v3.sqlobject.customizer.Bind
+import org.jdbi.v3.sqlobject.kotlin.attach
+import org.jdbi.v3.sqlobject.statement.SqlQuery
+import org.jdbi.v3.sqlobject.statement.SqlUpdate
+import org.springframework.stereotype.Component
+
+data class OnboardingSession(
```
vonElfvin, line 12: I see `result_member_id` in the migration but not here? 🤔

bystam: It's not used yet. That'll happen in a future PR. Since it's nullable we can also omit it from the model.

PR #298, opened 2021-09-10
feat/limit-start-date

fredriklagerblad:
# Jira Issue: [IPC-227] 

## What?
- Start validating quotes's startDate
- Add a general guidelines that is run for all quotes
- Add startDate validation as a general guideline, not allowing it to be more than 2 years ahead
- To cover the Rapio case, where the startData is only sent in the sign request (where the guidelines aren't run), we add an excplitit validation there
- For rapio cases, we also validate that the startDate is not in the past  

## Why?
- Don't allow startDates to far in the future
- For external quotes (Rapio), we don't won't to allow startDates in the past

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [ ] Tested locally

[IPC-227]: https://hedvig.atlassian.net/browse/IPC-227


src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
```diff
@@ -317,6 +318,9 @@ class SignServiceImpl(
                 if (startDate == null && currentInsurer == null) {
                     throw ErrorException(ErrorCodes.INVALID_STATE, "currentInsurer is required when startDate is null")
                 }
+                if (!it.isValidStartDate()) {
+                    throw ErrorException(ErrorCodes.INVALID_STATE, "Invalid startDate, too old or too far ahead in the future")
```
Michael-Duivestein, line 322: It would appear that this exception will be thrown under three different conditions:

1. No start date
2. Start date is more than 2 years into the future
3. Start date is in the past.

Can we maybe change the error description so that it represents more than one of the three conditions?

fredriklagerblad: No startDate should not trigger this exception, i.e. not considered invalid by the isValidStartDate() method.
(But no startDate and no currentInsurer would trigger, as it was before)

fredriklagerblad: And the message covers the 2 other cases.

Michael-Duivestein: Oops, didn't read the code properly. Sorry!

src/main/kotlin/com/hedvig/underwriter/service/guidelines/GeneralGuidelines.kt
```diff
@@ -0,0 +1,27 @@
+package com.hedvig.underwriter.service.guidelines
+
+import com.hedvig.underwriter.model.Quote
+import com.hedvig.underwriter.model.QuoteData
+import com.hedvig.underwriter.model.isValidStartDate
+import com.hedvig.underwriter.service.guidelines.BreachedGuidelinesCodes.INVALID_STARTDATE
+import com.hedvig.underwriter.service.guidelines.BreachedGuidelinesCodes.OK
+
+// These validate general things, not directly connected to a specific type
+object GeneralGuidelines {
```
meletios, line 10: Sweet

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/dtos/DebtDto.kt
```diff
@@ -1,6 +1,6 @@
 package com.hedvig.underwriter.serviceIntegration.memberService.dtos
 
-class DebtDto(
+data class DebtDto(
```
meletios, line 3: Is this related to this PR? 😅

meletios (approved):
Sweet, i have a naming suggestion

src/main/kotlin/com/hedvig/underwriter/model/Quote.kt
```diff
@@ -80,6 +80,25 @@ val Quote.danishTravel
 val Quote.validTo
     get() = this.createdAt.plusSeconds(this.validity)!!
 
+fun Quote.isValidStartDate(): Boolean {
```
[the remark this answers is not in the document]
fredriklagerblad: 👍 I fix.

src/main/kotlin/com/hedvig/underwriter/serviceIntegration/memberService/dtos/DebtDto.kt
fredriklagerblad: Sort of yes :-) 
Discovered when testing that the toString method of this class wasn't informational, since ot wasn't a data class ( like it should be)

PR #301, opened 2021-09-14
Feature/cross sell logic

cpiehl1:
# Jira Issue: [MX-274] 

## What?
- Move cross-selling logic from giraffe to underwriter

## Why?
- 

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [x] Tested locally

[MX-274]: https://hedvig.atlassian.net/browse/MX-274


src/main/kotlin/com/hedvig/underwriter/service/CrossSellServiceImpl.kt
```diff
@@ -0,0 +1,88 @@
...
+
+    private fun getSwedishCrossSells(contractTypes: List<TypeOfContract>): List<String> {
+        val isStudent = contractTypes.all { it.isStudentOrYouth() }
+        if (contractTypes.size == 1 && contractTypes.first().isHomeContent()) {
+            return if (isStudent) {
+                listOf("SE_ACCIDENT_STUDENT")
+            } else {
+                listOf("SE_ACCIDENT")
+            }
+        }
```
mightyost, line 32: I think it is better and clearer if to switch on contract type directly and not have `isHome()` and `isStudentOrYouth()` with its a bit funky logic...

I.e. something like this instead:
```
private fun getSwedishCrossSells(contractTypes: List<TypeOfContract>): List<String> {
        if (contractTypes.size != 1) {
            return listOf()
        }

        when (contractTypes.first()) {
            SE_HOUSE,
            SE_APARTMENT_BRF,
            SE_APARTMENT_RENT -> listOf("SE_ACCIDENT")
            SE_APARTMENT_STUDENT_BRF,
            SE_APARTMENT_STUDENT_RENT -> listOf("SE_ACCIDENT_STUDENT")
            else -> listOf()
        }
}
```

mightyost: ...same goes for the other markets...

src/main/kotlin/com/hedvig/underwriter/web/ExternalCrossSellController.kt
```diff
@@ -0,0 +1,28 @@
...
+
+@RestController
+@RequestMapping("/crossSells")
+class ExternalCrossSellController(
+    val crossSellService: CrossSellService
+) {
+    @GetMapping
+    @LogCall
+    fun getCrossSells(
+        @RequestHeader("hedvig.token") memberId: String
```
mightyost, line 20: Nice, this solves the auth problematics!

mightyost (approved):
Nice job, some cosmetic comments!

src/main/kotlin/com/hedvig/underwriter/service/CrossSellServiceImpl.kt
cpiehl1: Good point!

cpiehl1: It still became a bit funky with Denmark, where there can be two active contracts, but I think it looks better now 😃

src/main/kotlin/com/hedvig/underwriter/service/CrossSellServiceImpl.kt
```diff
@@ -0,0 +1,94 @@
...
+
+        return listOf()
+    }
+
+    private fun Contract.isActiveIsh(): Boolean {
+        val activeIshStatus = setOf(
+            ContractStatus.ACTIVE,
+            ContractStatus.ACTIVE_IN_FUTURE,
+            ContractStatus.ACTIVE_IN_FUTURE_AND_TERMINATED_IN_FUTURE,
+            ContractStatus.TERMINATED_TODAY,
```
meletios, line 88: Is this Activish?

meletios: 😄

cpiehl1: This status was included as `ActiveIsh` in the moving flow, so I felt it made sense to use the same logic.

PR #302, opened 2021-09-14
Remove deprecated field price from CompleteQuote

tobiasbexelius:
It has been deprecated since 2020-03-23 and was requested 0 times the last week [according to apollo](https://studio.apollographql.com/graph/Giraffe/fields?range=lastWeek&variant=current#CompleteQuote).


mightyost (approved):
👍

PR #306, opened 2021-09-17
Handle unknown insurers without failing the requests

fredriklagerblad:
# Jira Issue: [] 

## What?
- Map unknown insurers to Other

## Why?
- Via Rapio, people can send any string as currentInsurer, which then fails any GraphQL call tryting to load up the Quote.

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [ ] Tested locally


meletios (approved):

vonElfvin (commented):
What's the scenario where someone would send in some random input like this? Don't we have control of what is selected? Or is it a free-text form?

fredriklagerblad:
> What's the scenario where someone would send in some random input like this? Don't we have control of what is selected? Or is it a free-text form?

It's a String in the createQuote APIs, so clients can send in whatever. in the web onboarding it only sends valid ones (from the list we show them), but I can see random ones coming from mainly Rapio, but also IOS and Android.
Some examples: (Especially like the "Buffel & båg ab" one 😄 ) 
```
Har via resturang facket
Hedvig (via min sambo)
hedwig
jag och min tidigare sambo som du gått skilda vägar hade Hedvig men det var på hans personnummer
lärarförsäkring
redo
"Swedbanks egna, vet ej vad den heter"
Vet ej
Vill inte säga
water circles
Vet ej
villaägarna
Jag vill flytta min försäkring som jag har hos er
Komunal
Annat
Buffel & båg ab
har ingen
No
Hedvig but I can't can't join you're doesn't work!!!!
Hedvig fast inte jag som står på det
Hedvig men via inneboende
Jag har ingen men ”nej” fungerade ej
```

PR #308, opened 2021-09-20
Do not skip "reuse old price" logic, when the new price only contains a single lineitem

fredriklagerblad:
# Jira Issue: [] 

## What?
- Reuse old price when only one line item is present.

## Why?
- Now when changing price engine in SE, we started returning line items and that triggered a feature where we skpi the reusing of old prices.

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [ ] Tested locally


meletios (approved):
🚀

PR #310, opened 2021-09-21
Centralized validation of quote and active contracts combos for a mem…

mightyost:
# Jira Issue: [IPC-247] 

## What?

- Centralized validation of quote and active contracts combos for a mem…ber.

- Moved validation from sign strategy (validation has nothing to do with it) into the `QuoteBundleService` where it is more logical to be.

- Added validations for quote state, type families and valid bundling combos.

- Added concept of addon which is a type that requires another insurance to be signed.

- Cross compare quotes with existing contracts for member and make sure its a valid setup.

- Cross-sell logic is reusing bundle validation setup in `QuoteBundlingService`.

- A lot of test cases is fixed to comply with the new validation!

## Why?
- Need to support proper bundling and also incorporate active contracts.

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [ ] Tested locally

[IPC-247]: https://hedvig.atlassian.net/browse/IPC-247


src/test/kotlin/com/hedvig/underwriter/service/BundleQuoteServiceImplTest.kt
```diff
@@ -253,4 +305,107 @@ class BundleQuoteServiceImplTest {
...
+        every {
+            quote.market
+        } returns market
+
+        every {
+            quote.state
+        } returns QuoteState.QUOTED
+
+        return quote
+    }
```
fredriklagerblad, line 410: Maybe also a test of that non-QUOTED quotes cannot be bundled?

fredriklagerblad (approved):
Really nice! Just few small comments.

PR #314, opened 2021-09-24
Fix the missing switcher message + add quote Ids for traceability and better error messages

bystam:
# Jira Issue: [] 

## What?
- `userVisibleMessage` was missing
- Added quotes to the "get onboarding session" call, to make it easier to use/test
- added better error messages

## Why?
- This came up when testing it for real :) 

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [x] Tested locally


src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
```diff
@@ -49,19 +62,21 @@ class OnboardingService(
         val personalNumber = quotes.map { it.ssn }.toSet().single()
 
         if (memberService.checkIsSsnAlreadySignedMemberEntity(personalNumber).ssnAlreadySignedMember) {
```
bystam, line 64: Could've sworn I renamed this method in the other PR before I merged 🤔 but that must've been lost :(

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
```diff
@@ -25,6 +26,14 @@ class OnboardingService(
     private val apiGatewayService: ApiGatewayService
 ) {
 
+    companion object {
+        private const val SWITCHER_MESSAGE = "Jag har tagit del av förköpsinformation och villkor och bekräftar " +
```
tobiasbexelius, line 30: Should this message really be here, and only in Swedish?

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
```diff
@@ -33,35 +42,41 @@ class OnboardingService(
...
         bundleQuotesService.validateBundling(memberId = null, quotes = quotes)
 
         // Quotes are assumed to have a uniform market
         val market = quotes.map { it.market }.toSet().single()
         // Quotes are assumed to have a single personal number
         val personalNumber = quotes.map { it.ssn }.toSet().single()
 
-        if (memberService.checkIsSsnAlreadySignedMemberEntity(personalNumber).ssnAlreadySignedMember) {
-            throw ResponseStatusException(HttpStatus.FORBIDDEN)
+        if (memberService.isSsnAlreadySigned(personalNumber).ssnAlreadySignedMember) {
```
tobiasbexelius, line 64: 😌

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
bystam: It's only in Swedish in member-service, and there it's a constant in `application.yml`. Since it's same for all the profiles, I think it's even better left as just a constant.

tobiasbexelius: Ok, I'd say it's better to just have one copy of the same message though, don't you think?

bystam: What do you mean? The other copy is in member-service, which we will remove once we make the clients consume this API instead.

I modeled the new Swedish BankID API inside member-service to take the `userVisibleMessage` injected instead of taking some domain-specific boolean like `isSwitcher`, because that would've broken the abstraction that the API was de-coupled from onboarding in member-service.

tobiasbexelius: Sorry, I thought the copy application.yml was in this service too :)

tobiasbexelius (approved):

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
bystam: Ah! :) No it's only over there

PR #315, opened 2021-09-24
Use a callback url that is not escaped by FeignClients later

bystam:
# Jira Issue: [] 

## What?
- Use a callback URL structure that does not contain a query param
- Also make notification service optional, as failing to post to it shouldn't halt member creation

## Why?
- FeignClients escaped them, this was the quickest and easiest solution

image

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


src/main/kotlin/com/hedvig/underwriter/web/OnboardingSessionController.kt
```diff
@@ -100,26 +100,26 @@ class OnboardingSessionController(
...
     private val service: OnboardingService
 ) {
 
     companion object {
-        fun url(sessionId: UUID) = "/_/onboarding-sessions-se-bankid-callback?sessionId=$sessionId"
+        fun url(sessionId: UUID) = "/_/onboarding-sessions/$sessionId/se-bankid-callback"
     }
 
-    @PostMapping
+    @PostMapping("{id}/se-bankid-callback")
```
vonElfvin, line 112: I preferred the sessionId name :)
```suggestion
    @PostMapping("{sessionId/se-bankid-callback")
```

vonElfvin (approved):

src/main/kotlin/com/hedvig/underwriter/web/OnboardingSessionController.kt
bystam: I just like using the `id` name because it highlights that it's the primary ID of the resource presented by this controller. If I give it more custom names, to me that signals that it's some secondary ID.

These are the `OnboardingSession` controllers, so in my mind `id` is the less confusing, as I would've guessed `sessionId` refers to some query parameter foreign key or something.

PR #316, opened 2021-09-27
MX-297: redeem onboarding campaign codes upon signing

bystam:
# Jira Issue: [MX-297] 

## What?
- Redeem campaign code in onboarding sessions upon signing
- Also add a `timezone` property to the onboarding so that we can keep track of that properly (later we might even be able to flush it to other places?)

## Why?
- It's... necessary :D 

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [x] Tested locally

[MX-297]: https://hedvig.atlassian.net/browse/MX-297


src/main/kotlin/com/hedvig/underwriter/config/JdbiConfig.kt
```diff
@@ -27,3 +28,4 @@ fun Jdbi.install(): Jdbi = this.installPlugin(PostgresPlugin())
     .installPlugin(KotlinSqlObjectPlugin())
     .installPlugin(SqlObjectPlugin())
     .installPlugin(Jackson2Plugin())
+    .registerColumnMapper(JavaTimeMapperFactory())
```
bystam, line 31: Now we can have java time types in output beans :) (we could already INSERT them, that's in JDBI by default for some reason)

tobiasbexelius (approved):
Nice!

PR #319, opened 2021-09-29
Add campaigns to onboarding session

tobiasbexelius:
# Jira Issue: [MX-288] 

## What?
- Add campaign to onboarding session
- Add endpoints to add and remove campaigns

## Why?
- So that we can handle campaigns for memberless onboarding

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [x] Tested locally

[MX-288]: https://hedvig.atlassian.net/browse/MX-288


src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
```diff
@@ -132,6 +160,47 @@ class OnboardingService(
...
+            message = "Session with id: $onboardingSessionId was not found."
+        )
+
+        if (onboarding.campaignCode == campaignCode) {
+            repository.save(onboarding.copy(campaignCode = null))
+        } else {
+            throw notFound(
+                code = "campaign_code_not_found",
+                message = "Campaign code $campaignCode does not exist on onboarding session $onboardingSessionId"
+            )
```
bystam, line 175: This is not a big deal, but technically this is not a 404, but rather a bad request or a forbidden imo. 404 is meant to highlight that the requested URL did not reach a known destination. Usually that's reserved for either:
- There was no endpoint like that
- You send a "primary resource id" in the path that does not match an actual resource (as in `/thingy/{id}` where the ID is wrong)

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
```diff
@@ -132,6 +160,47 @@ class OnboardingService(
...
+        }
+
+        return RemoveCampaignResponse(
+            getBundledCost(
+                onboardingSessionId = onboardingSessionId,
+                campaignCode = null
+            ).price
```
bystam, line 182: Do we need to return the new price after removing? Wondering if it's just easier to include the price when fetching bundle on the onboarding session, and let the client just re-fetch the state upon mutation 🤔 then the client needs to make fewer assumptions about the state after changing something.

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
```diff
@@ -118,6 +128,24 @@ class OnboardingService(
...
+        if (session.campaignCode != null) {
+            throw forbidden(
+                code = "campaign_already_added",
+                message = "A campaign is already added to this onboarding session."
+            )
+        }
+
+        val bundledCostResponse = getBundledCost(onboardingSessionId, campaignCode)
+        repository.save(session.copy(campaignCode = campaignCode))
+        return AddCampaignResponse(bundledCostResponse.campaign, bundledCostResponse.price)
```
bystam, line 146: Same comment as downstairs. Would it be simpler to just answer "200 yep that code was good", and then have the client re-fetch the bundle and receive the cost that way:

https://github.com/HedvigInsurance/rfc/pull/5/files#diff-1171e0b69bce15da524c34c82b884bff7fcc35416db8ce6d2651e116a6898749R101

src/main/kotlin/com/hedvig/underwriter/web/OnboardingSessionController.kt
```diff
@@ -68,10 +71,24 @@ class OnboardingSessionController(
...
+    fun addCampaignToOnboarding(
+        @PathVariable id: UUID,
+        @RequestBody addCampaignDto: AddCampaignDto
+    ): AddCampaignResponse = service.addCampaign(id, addCampaignDto.code)
+
+    @DeleteMapping("/{onboardingSessionId}/campaigns/{campaignCode}")
+    fun removeCampaignFromOnboarding(
+        @PathVariable onboardingSessionId: UUID,
+        @PathVariable campaignCode: String
+    ): RemoveCampaignResponse = service.removeCampaign(onboardingSessionId, campaignCode)
```
bystam, line 88: Did you consider writing these as something like:

```http
// add a campaign
PATCH /{id}
{
  "campaignCode": "ABC123" 
}

// remove a campaign
PATCH /{id}
{
  "campaignCode": null 
}
```

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
```diff
@@ -118,6 +128,24 @@ class OnboardingService(
...
+        )
+
+        if (session.campaignCode != null) {
+            throw forbidden(
+                code = "campaign_already_added",
+                message = "A campaign is already added to this onboarding session."
+            )
+        }
+
+        val bundledCostResponse = getBundledCost(onboardingSessionId, campaignCode)
```
bystam, line 144: Delegating "add campaign" to "getBundledCost", and implicitly failing based on the errors of `ProductPricingService` to me seems like an unexpected codepath.

I personally would've expected something as simple as this:
```kotlin
fun setCampaignCode(onboardingSessionId: UUID, campaignCode: String?) {
  val session = repository.fetch(onboardingSessionId) ?: throw notFound()
  val market = ... // market from quotes somehow
  if (campaignCode != null) {
    if (!productPricingService.isValidCampaignCode(campaignCode, market)) {
      throw badRequest("invalid_campaign_code")
    }
  }
  // after saving, fetching the price for this onboarding session through its quotes should give the new price
  repository.save(session.copy(campaignCode = campaignCode))
}
```

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
bystam: We could easily "have both" in GraphQL by making the mutations return the entire onboarding session again, where we finish the mutation resolvers by just `return await upstream.underwriter.getOnboardingSession(id)` kinda

bystam (commented):
I think this is neat, but in my mind it could be a bit simpler. I wrote some thoughts about it.

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
bystam: I take forbidden back. I guess the most correct one would be 422 Unprocessable Entity https://stackoverflow.com/a/42171674

tobiasbexelius: 👍

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
tobiasbexelius: I'm not sure what is idiomatic graphql regarding the return type. Not sure what you mean about assuming state though, do you mean the onboarding session object itself?

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
tobiasbexelius: Possibly! Maybe we should iterate on this when adding the new quotebundle query?

src/main/kotlin/com/hedvig/underwriter/web/OnboardingSessionController.kt
tobiasbexelius: Hmm no. My initial thought on that is that it would add a little more complexity considering that we don't allow changing any other properties on the onboarding session

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
tobiasbexelius: Yeah, I'm not too happy with this part. How to refactor it will depend on what the input/output should be

src/main/kotlin/com/hedvig/underwriter/service/OnboardingService.kt
tobiasbexelius: I'll revert to InsuranceCost just in case 🤷‍♂️

bystam (approved):

PR #320, opened 2021-09-29
Change cross-selling logic and also return current co-insured

cpiehl1:
# Jira Issue: [] 

## What?
- Changed the cross-selling logic, determining active contracts is now split for base and addon contracts
  - For base, we only cross-sell when there exists a base contract without termination date
  - For addon, we only cross-sell when there is no contract, or it has a termination date before today
- Removed `CrossSellService`
- Added current number of co-insured to the response when asking for cross-sells

## Why?
- We split the logic for base and addon because for the base contract, we consider it to be active when no termination date is set, while for the addon, we consider it to be active if it has no termination date OR if the termination date is today or later.
- All the cross-selling logic is now in `BundleQuotesService`
- This will be used as default value in the embark flow

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [ ] Tested locally


mightyost (approved):
LGTM

PR #322, opened 2021-10-01
Generic: Introduced a new generic insurance quote object

mightyost:
# Jira Issue: [IPC-287] 

## What?
This is the initial commit for supporting the new generic flow where a new insurance can be added in a more streamlined fashion with minimal hardcoding.

- Added new type `Insurance` to define the information elements of future insurances.
- Added a new quote type to host the new insurances
- Added support for storing new quote type in db. Needed to add `InsuranceArgumentFactory` and `InsuranceDataMapper` for serialization/deserialization of quote into table.
- Added new data type `GenericData` that will contain the union of all existing and future insurance information elements. It will be used when to communicate with clients, PE and PP.
- Changed behaviour of `PersonPolicyHolder`, made `ssn`, `firstname`, `lastname` and `email` mutable (changing to `var`) and removed the `updateXxxx` methods. The old way did not work well for the new `QuoteData.InsuranceData`.

- Test for writing and reading new quote to db.

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [ ] Tested locally


src/main/kotlin/com/hedvig/underwriter/insurances/model/Insurance.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.neovisionaries.i18n.CountryCode
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
+@JsonSubTypes(
+    JsonSubTypes.Type(value = SwedishTestInsurance::class, name = "SwedishAccident")
+)
+interface Insurance {
+
+    @get:JsonIgnore
+    val name: String get() = javaClass.simpleName // Derived name must match @JsonSubTypes.Type above
```
bystam, line 17: I couldn't see where this property is used. Is it really practical to have it implemented like this rather than spelling it out explicitly in the `generic` property for each subclass? I mean the comment here actually isn't true for the only case right now :) ("SwedishTestInsurance" != "SwedishAccident").

If we do want the auto-implemented utility, and even enforce that it's not overriden, we could build it like this:
```kotlin
// top level, outside of the `Insurance` interface:
val Insurance.name: String get() = {
  // apply thread-safe cache or something if we're nervous about runtime data lookup performance
  val subtypes = Insurance::class.findAnnotation<JsonSubTypes>()!!
  val subtype = subtypes.value.first { it.value == this::class }
  return subtype.name
}
```

Then it will always match exactly what's written in the list above :) and since it's an extension function it can't be overridden.

src/main/kotlin/com/hedvig/underwriter/insurances/model/Insurance.kt
```diff
@@ -0,0 +1,30 @@
...
+
+    @get:JsonIgnore
+    val contractType: TypeOfContract
+
+    @get:JsonIgnore
+    val market: CountryCode
```
bystam, line 23: Wondering if it would be nice to group some of the values together that are going to be static values in the subclass as some kind of `InsuranceMetadata` type:
```kotlin
data class InsuranceMetadata(
  val market: CountryCode,
  val contractType: TypeOfContract
)
```

Then we would kind of group non-dynamic things like constants away from dynamic ones (like address)

src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
```diff
@@ -844,3 +886,41 @@ interface QuoteDao {
     )
     fun deleteLineItems(@Bind revisionId: Int)
 }
+
+internal class InsuranceDataMapper(val objectMapper: ObjectMapper) : RowMapper<InsuranceData?> {
```
bystam, line 890: Is there anything happening in this class that's custom outside of the `val insuranceObject = objectMapper.readValue(insuranceObjectAsJson, Insurance::class.java)`? If not, then you could also just have implemented a `ColumnMapper<Insurance>` and leave everything else as-is. At a glance, it looks like all the other things should work out of the box.

src/main/kotlin/com/hedvig/underwriter/model/QuoteDao.kt
```diff
@@ -844,3 +886,41 @@ interface QuoteDao {
...
+
+        val internalId = r.getInt("internal_id")
+        val id = r.getObject("id", UUID::class.java)
+        val typeOfContract: String = r.getString("type_of_contract")
+        val insuranceName: String = r.getString("insurance_name")
+        val market = r.getString("market")
+        val ssn: String? = r.getString("holder_ssn")
+        val firstName: String? = r.getString("holder_firstname")
+        val lastName: String? = r.getString("holder_lastname")
+        val birthdate: Date? = r.getObject("holder_birthdate", Date::class.java)
```
bystam, line 902: I think LocalDate should work with JDBI 🤔

src/main/kotlin/com/hedvig/underwriter/service/SignServiceImpl.kt
```diff
@@ -555,7 +558,15 @@ private fun updateSsnFromRequest(
         throw IllegalArgumentException("Invalid ssn, does not match existing ssn in quote")
     }
 
-    return quote.copy(data = quote.data.updateSsn(ssn = ssn))
+    when (quote.market) {
+        Market.NORWAY -> require(ssn.isValidNorwegianSsn()) { "Illegal SSN" }
+        Market.DENMARK -> require(ssn.isValidDanishSsn()) { "Illegal SSN" }
+        Market.SWEDEN -> "We should add validator for SE personnummer"
```
bystam, line 564: I built one in member-service ^^

src/main/kotlin/com/hedvig/underwriter/config/JdbiConfig.kt
```diff
@@ -16,11 +18,18 @@ import org.springframework.context.annotation.Configuration
...
+        val jdbi = Jdbi.create(dataSource).install()
             .apply {
                 getConfig(Jackson2Config::class.java).mapper = objectMapper
             }
+
+        // Globally register row and argument mappers for insurance quotes
+        jdbi.registerArgument(InsuranceArgumentFactory(objectMapper))
+        jdbi.registerRowMapper(InsuranceDataMapper(objectMapper))
+
+        return jdbi
```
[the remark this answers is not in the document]
bystam: I realize now that it's because `install()` doesn't have access to an object mapper.

Honestly I think the real solution is to require that it does, and then change the `JdbiRule` to create an object mapper as well. Then all other tests won't need to fight with this.

bystam (commented):
Woop woop! Added some thoughts and questions :)

src/main/kotlin/com/hedvig/underwriter/insurances/model/GenericData.kt
```diff
@@ -0,0 +1,46 @@
...
+    val student: Boolean? = null,
+    val youth: Boolean? = null
+) {
+
+    data class ExtraBuildingRequestDto(
+        val id: UUID?,
+        val type: String,
+        val area: Int,
+        val hasWaterConnected: Boolean
+    )
```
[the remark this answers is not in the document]
bystam: I still think the suffix `Request` feels a bit out of place. It's not really a request? It's just an object on the generic data 🤔

src/test/kotlin/com/hedvig/underwriter/testhelp/QuoteBody.kt
```diff
@@ -229,4 +229,25 @@ sealed class QuoteBody(
             "student" to isStudent
         )
     }
+
+    class SwedishAccident2 : QuoteBody(
+        ssn = "199110112399",
+        birthDate = LocalDate.of(1944, 8, 4),
```
bystam, line 235: The ssn and birthdate don't match here. That usually makes tests fail for me.

PR #323, opened 2021-10-01
MX-300, MX-303: Implement quote subset signing and SE BankID progress reporting

bystam:
# Jira Issue: [MX-300] [MX-303] 

## What?
- Replace the 'sign_status' column with a sub-entity on `OnboardingSession` called `signing`. A nullable property that tracks the current signing state
- The previous `OnboardingSessionSignStatus.NONE` is now just `signing == null`.
- This signing object also keeps track of quote IDs actively being signed, to support signing only a subset of the quotes in the session (this is optional, if omitted all of them are signed)
- The signing object also keeps track of the most recently reported `hintCode` from the BankID API

## Why?
- These were 2 feature requests that came up during the GraphQL API review:
  - We need feedback on the most recent bankid status (hint code)
  - We need to be able to sign a subset of quotes (SE accident flow)

## Optional checklist
- [x] Codescouted
- [x] Unit tests written
- [ ] Tested locally

[MX-300]: https://hedvig.atlassian.net/browse/MX-300
[MX-303]: https://hedvig.atlassian.net/browse/MX-303


src/test/kotlin/com/hedvig/underwriter/web/OnboardingSessionControllerTest.kt
```diff
@@ -262,6 +287,52 @@ internal class OnboardingSessionControllerTest : IntegrationTest() {
...
+                productPricingClient.createContract(capture(slot), any())
+            }
+        }.captured
+
+        assertThat(contractsRequest.quotes.size).isEqualTo(1)
+        assertThat(contractsRequest.quotes.single().quoteId).isEqualTo(apartment.id)
+    }
+
+    @Test
+    fun `SE quote - bankid callback is idempotent`() {
```
bystam, line 319: The member-service API of course shouldn't fire more than once.. but you never know :) if the call to underwriter times out, this is really nice to have.

src/main/resources/db/changelog/migrations/20210930-1700-add-onboarding-signing.sql
```diff
@@ -0,0 +1,22 @@
...
+);
+
+CREATE TABLE onboarding_session_signing_quote_links(
+    session_id UUID NOT NULL,
+    quote_id UUID NOT NULL,
+    PRIMARY KEY (session_id, quote_id)
+);
+
+ALTER TABLE onboarding_sessions
+ DROP COLUMN sign_status;
```
bystam, line 18: Dropping a column is safe because nobody is using this anyway

tobiasbexelius (approved):

PR #326, opened 2021-10-05
Generify and consolidate validation of SSN and setting birthdate from it.

mightyost:
Validating SSN and setting a birthdate from it was very specific to the insurance type and shattered all over the codebase.

With this commit it is generified (using market rather than the specific insurance type) and consolidated to the central places where needed.

## Optional checklist
- [ ] Codescouted
- [x] Unit tests written
- [ ] Tested locally


Michael-Duivestein (approved):
nice cleanup

PR #327, opened 2021-10-05
Remove service interfaces

tobiasbexelius:
## What?
- Remove interfaces for services (and a couple of repositories) that just has one implementation suffixed with Impl

## Why?
- To reduce mental and visual clutter


bystam (approved):
🎉

PR #333, opened 2021-10-07
Validate if already signed using the ssn instead of the memberid

fredriklagerblad:
# Jira Issue: [] 

## What?
- Validate if already signed using the ssn instead of the memberid

## Why?
- Since a new member might not have a ssn associated with it yet.

## Optional checklist
- [ ] Codescouted
- [ ] Unit tests written
- [ ] Tested locally


AlexSJ1990 (approved):
✅

